PR #4, opened 2015-10-15
Added contextual message

jvt:
Context messages are slightly smaller than the "x / 4 pull requests submitted" string.

### DIFF
--- a/views/index.hbs
+++ b/views/index.hbs
@@ -36,9 +36,9 @@
         {{#if prs}}
             <div class="row">
                 <div class="large-12 columns">
-                    <h2><strong>{{ prs.length }} / 4</strong> pull requests submitted in Hacktober.</h2>
+                    <h2><strong>{{ prs.length }} / 4</strong></h2>
                     <br>
-                    <h3 class="center"><strong>{{ statement }}</strong></h3>
+                    <h3><strong>{{ statement }}</strong></h3>
                 </div>
             </div>
 
--- a/views/index.hbs
+++ b/views/index.hbs
@@ -36,7 +36,9 @@
         {{#if prs}}
             <div class="row">
                 <div class="large-12 columns">
-                    <h2><strong>{{ prs.length }} / 4</strong> pull requests submitted in Hacktober</h2>
+                    <h2><strong>{{ prs.length }} / 4</strong> pull requests submitted in Hacktober.</h2>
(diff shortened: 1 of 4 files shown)


views/index.hbs
```diff
@@ -36,7 +36,9 @@
         {{#if prs}}
             <div class="row">
                 <div class="large-12 columns">
-                    <h2><strong>{{ prs.length }} / 4</strong> pull requests submitted in Hacktober</h2>
+                    <h2><strong>{{ prs.length }} / 4</strong> pull requests submitted in Hacktober.</h2>
+                    <br>
+                    <h3 class="center"><strong>{{ statement }}</strong></h3>
```
jenkoian: Can get rid of the class here as not used.

views/index.hbs
```diff
@@ -36,7 +36,9 @@
         {{#if prs}}
             <div class="row">
                 <div class="large-12 columns">
-                    <h2><strong>{{ prs.length }} / 4</strong> pull requests submitted in Hacktober</h2>
+                    <h2><strong>{{ prs.length }} / 4</strong> pull requests submitted in Hacktober.</h2>
```
jenkoian: I'm thinking we get rid of the 'pull requests submitted in Hacktober' bit now, what you think?

views/index.hbs
jenkoian: Can you also add this to the line below line 69 as well (I know the logic here isn't perfect, but from what I can tell you can't do much in terms of expressions in handlebars and didn't bother adding a custom helper or anything)

views/index.hbs
jvt: I think that would make the most sense to do.

views/index.hbs
```diff
@@ -36,7 +36,9 @@
         {{#if prs}}
             <div class="row">
                 <div class="large-12 columns">
-                    <h2><strong>{{ prs.length }} / 4</strong> pull requests submitted in Hacktober</h2>
+                    <h2><strong>{{ prs.length }} / 4</strong></h2>
+                    <br>
+                    <h3><strong>{{ statement }}</strong></h3>
```
jenkoian: Last thing, can you copy this line to just below line 69 too. Cheers!

jenkoian:
Thanks!

PR #9, opened 2015-10-19
Set min-width to 360px to prevent overlapping

evidex:


jenkoian:
Thanks for the PR! Can you just give a brief explanation of the issue?

evidex:
It's trivial really. When viewing https://hacktoberfestchecker.herokuapp.com/ on a screen of width < 360px, the title "Hacktoberfest Checker" overlaps the "Fork me on Github" ribbon. This PR sets the min-width to 360px in order to prevent this occurring on small screens.

evidex:
Looking more closely, there's another occurrence of this between 450px and 490px width. I've added an additional media query to correct that also. Like I said, it's only a nit-picking change :smile:

css/main.css
```diff
@@ -12,6 +12,7 @@
 
 body {
     display: block;
+    min-width: 360px;
```
jenkoian: I think most phone are 320px or 360px so maybe target 320px here? Currently with this at 320px it introduces side scrolling which looks a bit weird.

css/main.css
```diff
@@ -44,6 +45,12 @@ h1 {
     font-size: 7em;
 }
 
+@media screen and (max-width: 490px) {
+    h1 {
+        width: 93%;
```
jenkoian: Another options is we just reduce the font size? In fact we could just do that, below 490px reduce the font size to 2.5em or something? What you think?

evidex:
@jenkoian Solid suggestions. I've updated the PR to set min-width of 320px, and swapped it around a bit so that it's mobile first, with media queries to account for larger screens, rather than smaller. Finally, the overlap fixes use font-size instead, along with a little bit of padding.

evidex:
Realized I'd not corrected the h2/h3 sizes to adapt for the previous commit. Updated :smile:

css/main.css
```diff
@@ -41,23 +42,41 @@ h1, h2, h3 {
 
 h1 {
     font-family: squealerregular;
-    font-size: 7em;
+    font-size: 2.5em;
+    padding-top: 10%;
+}
+h2, h3 {
```
jenkoian: Minor thing but could you add newline between rules here.

css/main.css
```diff
@@ -41,23 +42,41 @@ h1, h2, h3 {
 
 h1 {
     font-family: squealerregular;
-    font-size: 7em;
+    font-size: 2.5em;
+    padding-top: 10%;
+}
```
jenkoian: Could you add newline here please

css/main.css
```diff
@@ -41,23 +42,41 @@ h1, h2, h3 {
...
     }
 }
 
-h2 {
-    font-size: 3em;
+@media screen and (min-width: 750px) {
+    h1 {
+        font-size: 5em;
+        padding-top: 4%;
+    }
```
jenkoian: Could you add newline here please

css/main.css
```diff
@@ -41,23 +42,41 @@ h1, h2, h3 {
...
+    h1 {
+        font-size: 6em;
+        padding-top: 2%;
+    }
 }
 
+
+
+
+
```
jenkoian: Can you remove this whitespace cheers.

jenkoian:
Looks great thanks, just a few minor things and should be good for merge then.

evidex:
Complete :smile:

jenkoian:
Thanks!

PR #14, opened 2015-10-30
Make get parameters work

andeersg:
I saw your todo "Fix ability to link to users that ajax fix above broke (e.g. http://localhost:5000/?username=jenkoian)" and I think this fixes it.

### DIFF
--- a/views/index.hbs
+++ b/views/index.hbs
@@ -33,7 +33,7 @@
             </div>
         </form>
 
-        <div id="results"></div>
+        <div id="results">{{> prs}}</div>
 
         <footer>
             <div class="large-12 columns text-center">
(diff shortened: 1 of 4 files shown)


jenkoian:
Awesome, thanks! One thing I noticed, it doesn't retain the username in the search bar. Also, would be *really* neat if when submitting via ajax it updated the URL also. :)

andeersg:
I updated the pull request with those features :)

jenkoian:
Awesome work thanks!

views/index.hbs
```diff
@@ -33,7 +33,7 @@
             </div>
         </form>
 
-        <div id="results"></div>
+        <div id="results">{{> prs}}</div>
```
jenkoian: Sorry, been trying to look into up but can't find it...mind sharing with me what the `>` does here? I know it's important but just wondering :)

jenkoian: Ah just found this http://handlebarsjs.com/partials.html :+1:

andeersg: Yeah, partials in Handlebars are pretty nice :smile:

Issue #23, opened 2016-09-30
Live hacktoberfest issue ticker


jenkoian:
Bit more challenging this one but think it'd be quite neat to have some kind of ticker somewhere which pops up with issues recently opened on github with the hacktoberfest label. 

You should be able to use https://developer.github.com/v3/search/#search-issues which you can filter by labels.

bmorelli25:
I might give this a go at somepoint in the next few days. I'm more familiar with react than jQuery, so no promises that I'll be successful, but worth a shot at least.

If all else fails, I'll at least write some tests per #16

mattjegan:
Perhaps you could use the GitHub API to get the issues list? The docs for it are [here](https://developer.github.com/v3/search/#search-issues)

Maybe something along the lines of this:
```
https://api.github.com/search/issues?q=label:hactoberfest&state:open&sort=created&order=asc
```
which should return all issues with the `hactoberfest` tag that are `open` sorted by when they were made.

hamzahsarwari:
Thank you to whoever made this site. It is incredibly amazing! <3

Ne3l:
Did my attempt to fix that (https://github.com/jenkoian/hacktoberfest-checker/pull/32)

The presentation of the issues can be much better also might be interesting letting the user filter the issues by some criteria as github such as programming language.

REPTILEHAUS:
Yeah this  is pretty cool 👍 if you need a hand with adding in any features or stuff let me know

jenkoian:
Thanks @REPTILEHAUS there is an ongoing PR [here](https://github.com/jenkoian/hacktoberfest-checker/pull/32) - feel free to get involved if you wish 👍

REPTILEHAUS:
Cool, Ive had a few whiskeys at the moment but Im actively taking part in Hacktoberfest so Ill certainly start looking into it tomorrow at 12pm

jenkoian:
If anyone is interested in helping make this 'live' there is a brilliant example of using sockets in a node app [here](https://github.com/debugger22/github-audio)

jenkoian:
Thanks to @Ne3l & @games647 hacktoberfest issues are now being brought out on the homepage. I still think it'd be cool to have this live update though so gonna leave this ticket open.

redice44:
I'd like to take a crack at the live update part, if it's still up for grabs. 👍

babhishek21:
[...]

PR #32, opened 2016-10-01
Live hacktoberfest issue ticker #23

Ne3l:
Finds issues with hacktoberfest label and display them below.

The presentation of issues can be improved also i'm not sure if using github label colors was a good idea.


jenkoian:
Thanks for this! Here's my feedback (mostly revolves around look and feel):

* Site uses [Foundation](http://foundation.zurb.com/) so please leverage some of that for the styles. For example there is a [label component](http://foundation.zurb.com/sites/docs/label.html) which you can use for displaying the labels. 

* I'd say display the issues in two columns underneath the search bar with a smaller font to make it a bit subtler. I also think we should display the issues from the get go rather than on button click. 

* I'd make the titles clickable rather than the description. Also, the labels should be clickable too. 

* Not sure whether to change it to descending created date or not. On one hand it'd be good to show the latest issues (then in future we could have some kind of real time check which could pop new issues to the top, maybe) but on the other hand I think it'd be neat if the old issues get addressed, so am up for opinions on this.

Ne3l:
Thanks for the feedback, i agree with the most but:

- If labels should be clickable what should the click action do?

- For display the issues i think it would be a good idea to dispose some kind of filters such a order asc/desc, language, labels, etc...

jenkoian:
Cheers @Ne3l and @games647 - it's definitely getting there but still feels a little clunky. 

* I think the avatars are too big at the mo and they should be round to match the avatar on the result of the check. 
* In general I think it should be a bit subtler, it's a secondary feature after the main feature which is the check, so I'd bring the size of the font and everything down a little. Particularly the title (the 'Help other projects - Total: 4065' bit). 
* Labels I think should go to the issues of the project filtered by the label you've clicked.
* With regards filtering, I'm not sure, think it'd be doing too much then. See the main benefit in this just to give a user a quick heads up of potential contributions, they can go through to github then to drilldown further.

jenkoian:
Looking good! I've made a few tweaks, mostly for the responsive stuff and pushed up to a branch here: https://github.com/jenkoian/hacktoberfest-checker/commit/67f87d4508840e6734d52774800fd08e649993de if you're happy with those changes, you can either open up your branch so I can write to it (e.g. https://github.com/blog/2247-improving-collaboration-with-forks) or I'll PR that branch I just pushed (it has your commits so your collaboration should remain in tact). Cheers!

jenkoian:
Merged in https://github.com/jenkoian/hacktoberfest-checker/pull/47 thanks for all your help!

PR #33, opened 2016-10-02
Add dockerfile to run app within docker

FuriKuri:
I added a dockerfile, so it's possbile to create a docker image and run the app within a docker container. So it's not necessary to have node installed.


jenkoian:
Thanks!

PR #35, opened 2016-10-03
Add a background image from official site

kumabotz:


jenkoian:
Thanks @kumabotz, as this app is not affiliated with Digital Ocean in any way I think we'd need to check with them before using their image though. If you can get approval, I'm happy to merge, otherwise I'm going to have close this I'm afraid.

kumabotz:
All good! 🎃

PR #44, opened 2016-10-05
Added Travis CI

MiguhRuiz:


.travis.yml
```diff
@@ -0,0 +1,6 @@
+language: node_js
+node_js:
+  - "6"
+script: npm run postinstall
+script: node index.js
```
jenkoian: How does this work in travis environment? As it just loads the server and waits for input, I'm not sure how travis knows when it's complete?

jenkoian:
Thanks for this @MiguhRuiz I think we should add some basic tasks like linting (using es lint or something). Also thinking it may be worth building against a few versions of node, perhaps current LTS (4.6 I think) and latest stable (6.7 I think).

.travis.yml
MiguhRuiz: Per default it runs `npm install` and then `npm test`but if you dont have any test you might configure something like I did.

jenkoian:

PR #50, opened 2016-10-07
Improve Dockerfile to cache node_modules

matteo-hertel:
There are a lots of way to improve the node usage with Docker, one of them is to cache the node_modules folder so that npm install runs only if the package.json changes

see 
[https://nodesource.com/blog/8-protips-to-start-killing-it-when-dockerizing-node-js/#protip3startcachingnode_modules](https://nodesource.com/blog/8-protips-to-start-killing-it-when-dockerizing-node-js/#protip3startcachingnode_modules) for more details

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,19 @@
 FROM node:6
 
 WORKDIR /app
-COPY . /app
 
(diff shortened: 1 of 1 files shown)


jenkoian:
Cheers!

Issue #16, opened 2016-09-28
Add tests


jenkoian:
This app is something I knocked together pretty quickly, but as I now seem to be recycling it every year it'd be cool to make it a bit more maintainable. So it'd be cool to get some tests in place so that if I did wanna refactor I could do say with a modicum of reassurance.

hamzahsarwari:
This works great for me! and i love it thank you :)

matteo-hertel:
Adding some tests it's a very good idea, however most of the code is one file (both front end and back end).

I believe the fist step will be split the code in reusable components, maybe using ES6 or TypeScript to have it more maintainable, what do you think?

jenkoian:
@matteo-hertel agree it probably needs a little refactoring to make it easily testable. Would like to keep the app simple though, it doesn't really do too much to warrant bringing in a mass of dependencies etc plus I think that part of the reason it attracts contributors is that's it's easy for beginners to work with. Having said that I'm not opposed to typescript or in particular es6 so long as we can keep the app simple. Another option for testing would be to have just end to end tests although would need to introduce some kind of stubbed GitHub api.

PR #53, opened 2016-10-08
Refactor the front end, components should be easier to test now

matteo-hertel:
After discussing some refactoring in #16 before adding tests I've decided to start from the front end that it's easier to de-couple.

I went for plain ol' JavaScript, no ES6/TypeScript so it will be easier for other developer to understand what's going on and there is no need to know ES6/TS

If/when we're all happy with this refactor I can start to add tests!

### DIFF
--- a/public/js/script.js
+++ b/public/js/script.js
@@ -1,28 +1,152 @@
 $(document).ready(function() {
-
-    $('input[name=username]').focus();
-
-    $('form').on('submit', function(e) {
-        var name = $('input[name=username]').val();
-
-        if(name == '') {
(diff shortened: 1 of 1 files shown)


jenkoian:
@matteo-hertel wow this looks great! Would like to checkout and test before merging and unfortunately won't be able to get to a computer until Monday - but I'm +1 with approach taken so far 👍

matteo-hertel:
@jenkoian no worries, have a look whenever you can, I might add some test just to see how it goes

jenkoian:
@matteo-hertel works as expected, thanks!

Issue #40, opened 2016-10-05
Add graph of PRs opened throughout Hacktoberfest


jenkoian:
Was thinking it'd be cool to have a graph or like a sparkline somewhere of PRs opened throughout October, maybe we can dip a little in to September also to give a comparison.

Should be able to call something like https://api.github.com/search/issues?q=created:2016-09-30T00:00:00-12:00..2016-10-01T23:59:59-12:00&sort=created&order=asc&type=pr which will give you the `total_count` at the top. We can then use this total count for each time period (each day I guess) and then plot our graph. If that ends up firing off too many API requests (it could potentially mean over 30 API requests per request) then possibly we can just give that one API request the complete range (e.g. mid september to end of october) and then traverse the results to work out when the PRs were created. Results are paginated though so would need to take that into account also.

darrylpargeter:
Hey @jenkoian 

Is this still open to be worked on?

jenkoian:
@darrylpargeter yep! :)

darrylpargeter:
Then I shall make a start, any preference on charting tools? or would I be free to use the one I know?

jenkoian:
No preference on charting tools, I really like the sparklines on the repositories view on GitHub e.g. https://github.com/jenkoian?tab=repositories althougj we would need a bit more info like an axis with labels I suspect but yeah generally up to you :)

darrylpargeter:
so a minimalist Line graph it is ^^, should it be displayed on the main page or on a different page?

jenkoian:
Was thinking the main page but could put it on another page if it doesn't seem to fit anywhere

darrylpargeter:
kool, will put it not the main page for now, then can go from there

jenkoian:
Sounds good 👍

PR #7, opened 2015-10-16
Reduced risk of reaching rate limit on github API

fdoxyz:
Implemented with `memory-cache`

PR #56, opened 2016-10-09
Add a graph that maps all of the pull requests though out October

darrylpargeter:
added a graph using D3 for issue #40

### DIFF
--- a/public/js/graph.js
+++ b/public/js/graph.js
@@ -1,0 +1,209 @@
+/*******************************************************
+ *                                                     *
+ *                GET GRAPH DATA                       *
+ *                                                     *
+ ******************************************************/
+
+getData(); // start getting the graph data (kind of slow atm)
+
+
+function getData() {
+
+var data = []; // set empty gobal array
+var currentDate = new Date;
+  
+  for ( var i = 1; i <= currentDate.getDate(); i++) {
+    var startDate = setStartDate(i)
+
+    d3.json('/graph?start='+startDate+'&end='+i)
+      .get(function(res) {
+        // create a global graphDate var
+        data.push(res);
+
+        if (i - 1 == data.length) {
+          // Sort the grid into date order
+          data.sort(function(a, b) {
+            return new Date(a.date) - new Date(b.date);
+          });
+
+          buildGraph(data);
+        }
+      });
+  }
+}
+
+function setStartDate(endDate) {
+  if (endDate == '01'){
+    startDate = 30;
+  } else {
+    startDate = endDate - 1;
+  }
+  return startDate  < 10 ? "0"+startDate : startDate;
+}
+
+
+/*******************************************************
+ *                                                     *
+ *                BUILD GRAPH DATA                     *
+ *                                                     *
+ ******************************************************/
+  var margin = {top: 20, right: 20, bottom: 30, left: 50},
+      width  = 960 - margin.left - margin.right,
+      height = 150 - margin.top - margin.bottom;
+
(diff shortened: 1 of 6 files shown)


games647:
d3 isn't present in the bower.json or am I missing something?

darrylpargeter:
it's not, I must have forgot to add the save flag, will fix that now

darrylpargeter:
That has now been added

public/js/graph.js
```diff
@@ -0,0 +1,209 @@
...
+ *                                                     *
+ ******************************************************/
+
+getData(); // start getting the graph data (kind of slow atm)
+
+
+function getData() {
+
+var data = []; // set empty gobal array
+var currentDate = new Date;
```
jenkoian: Indent these.

jenkoian:
Thanks @darrylpargeter this is absolutely fantastic! Few things to address then will get this merged 👍 

* Indentation - we've used 4 space tabs everywhere so far, can you update your code to use 4 spaces instead of 2? 
* I think we should have the start date be like the beginning of September and the end date be  like the env of November. We can then see easily how much of a difference hacktoberfest makes compared to other months, what do you reckon?

jenkoian:
Screenshot for reference:

screen shot 2016-10-10 at 10 50 25

darrylpargeter:
I Shall add the tab spacing.

I'm game for making the start date from the start of September, I will have to think of away around the rate limit (had a check and it is 30 requests per minutes for /search)

jenkoian:
Thans @darrylpargeter for the rate limit stuff, we can probably cache this quite heavily. We only really need to get the data once a day. Am thinking we could store the etag and save it for 24 hours or something, sending if none match headers. We did do this at one point, will try and dig out the commit.

jenkoian:
@darrylpargeter here you go, this kind of thing: https://github.com/jenkoian/hacktoberfest-checker/pull/7/files

darrylpargeter:
I shall have a look and come up with something

jenkoian:
@darrylpargeter if you address the tabbing we can look at the caching thing in a future PR (really keen to get this awesome feature out! :))

darrylpargeter:
@jenkoian I shall up date the pull once I get home after work

darrylpargeter:
Tab spacing should be fixed now

jenkoian:
@darrylpargeter this looks *exactly* like what I was hoping for. I'm going to hold off merging though until I can spare some time to look at the caching. The search API we're using has a rate limit of 30 requests per minute, and we're calling it 31 times (if I've read it correctly) to render the graph. Not sure there are any other solutions than doing it once and then caching for 24 hours, but if there are I'm all ears. Thanks again for your wonderful PR, determined to get this up on the site soon 👍

darrylpargeter:
Hey @jenkoian  sorry about the delay, but I was thinking it might be better to store the values in a json file (to save having to spin up a database) then to cache them every 24 hours as with caching we would still need to make the calls to the api which by the end of the month will take over 2 minutes. 
though if we save it all into a json file from the 1st September to the current date then when a user loads the site if the last date in the json file and current date don't match then we call the api to update the file then the next user won't have to call the api but just the json file and we shouldn't hit the rate limit this way.

what do you think?

jenkoian:
Hey @darrylpargeter I did a bit of digging into the caching thing I was talking about and it seems the search endpoint doesn't support etags (guess because of the volatile nature of the response) so my original idea won't work. So, I am very much in favour of your approach! Seems like a neat solution to me 👍

jenkoian:

PR #57, opened 2016-10-09
Open Graph meta data

peterkokot:
Hello,

this is a minor patch that adds some meta tags for Open Graph such as Title, description and image.

Image is a simple screenshot from the current website.

Thank you considering merging it to the repository. In case there needs to be some adjustments, let me know.


jenkoian:
Cheers!

Issue #51, opened 2016-10-08
Add language to the list of hacktoberfest repos


damiankw:
I would like to see a 'php', 'javascript' or whatever next to the list of hacktoberfest labelled issues underneath the score, to make it easier for newer guys to find projects they can work on in their language.

damiankw:
I attempted to look at this myself, before I realised it was node.js, that I knew nothing about node.js, and after spending two hours on it, couldn't figure out how to use the npm-github-api to gather any detail about the repo :P

matteo-hertel:
I was thinking on adding that, if the API call will return the languages used on the PR, it should be straightforward!

damiankw:
That's exactly what I was thinking, unfortunately the 'issues' search doesn't return the languages as far as I can see (at least, the npm-github-api doesn't). I assume you will have to use the API to pull the actual GitHub repo detail, which will have it in there .. but I couldn't figure out how to get that to work!

matteo-hertel:
The library does not have support for getting some repo infos, but is possible to call the REST API like

`https://api.github.com/repos/jenkoian/hacktoberfest-checker/languages`

and will return the languages for the given repo

is not that straightforward to add because for each PR you need to get the repo and everything is promised based so you need to add the call at the right moment of the flow to add the correct data for the languages

matteo-hertel:
This can now be closed, added the language in #58

PR #58, opened 2016-10-09
Add repo languages

matteo-hertel:
As discussed in #51 I've added the language to the repo list

now the first limitation is: the library used to interact with GitHub when calling the repo info is returning just one language (probably the most used) instead the API exposed by github returns all the languages used in the repo [https://developer.github.com/v3/repos/#list-languages](https://developer.github.com/v3/repos/#list-languages)

I think ATM will do fine, but the back-end need some restructure for both maintainability and testing

### DIFF
--- a/index.js
+++ b/index.js
@@ -113,7 +113,8 @@ function getIssues(){
 
         totalIssues = res.total_count;
 
-        _.each(res.items, function(issue) {
+        var issuesLen = res.items.length;
+        _.forEach(res.items, function(issue, k) {
             var issueUrl = issue.html_url;
 
             var repo_url = issueUrl.replace(/\/(issues)\/\d+/, "");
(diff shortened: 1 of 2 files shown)


jenkoian:
Thanks @matteo-hertel this looks good - are you able to resolve the conflict? Once that's done I'll check out locally to give it a quick check and then look to merge 👍

matteo-hertel:
Yes sure, I'll fix the conflict with the current master and I'll send the pull request back

matteo-hertel:
I've fixed the conflicts and fetch the latest version from upstream so it should be good to go now

jenkoian:
@matteo-hertel lovely thanks! One minor thing, I think it'd look a little nicer if we just had the language 'pill' next to the title of the. For example: 
screen shot 2016-10-11 at 10 59 02

What do you reckon?

matteo-hertel:
I think it's a very good idea, leave that with me and I'll update the view and the style

matteo-hertel:
Alright, I've updated the view, it looks so much better now!

jenkoian:
Awesome thanks @matteo-hertel

PR #61, opened 2016-10-10
Separates concerns into modules for easier testing

josecanhelp:
Separating each piece of the back end into multiple modules will allow us to pull them into individual tests.


jenkoian:
Thanks @josecanhelp I really like this refactoring. I'm wary it will likely break the other open PRs so would like to hold off on merging until those are closed off (they are close). Another thing I noticed you've used some ES6 syntax in places. If we're gonna target ES6 I think we should make this explicit (in README and possibly through an nvmrc or something) and also go full ES6, e.g. vars to lets etc. What do you reckon?

josecanhelp:
You make valid points @jenkoian. I did use ES6 syntax, but that can be modified to use ES5 syntax pretty easily. 

I like the idea of moving to full ES6, but that's totally your call. 

I understand this project is a great starting point for people to contribute to Hacktoberfest, so I don't want to make it too complicated for others to understand. At the same time, I don't know how else to properly test the codebase without a refactor.

Thanks for managing this repo. It's a fun project.

jenkoian:
@josecanhelp yeah I like your refactoring because it breaks it up nicely without adding complexity. As for ES6 I think stick to ES5 for this PR and then we can maybe open a new PR migrating to ES6. We can then easily point towards the canonical PR if people are wondering about where the switch came in.

PR #54, opened 2016-10-09
#52 added share buttons

Ferbla:
Added social media buttons for when a user is searched.

PR #62, opened 2016-10-10
Added Sharing Icons

123aswin123:
Hey Guys, I've added non tracking sharing icons 😄  You could use this!


jenkoian:
Thanks @123aswin123. Few things, the icons seem a little off centre for me (see screenshot below). Also I think the icons should appear only after user lookup and the text for the tweet/status/g+ whatever should be context aware too, like '1/4 PRs done for #hacktoberfest, check how you're doing at...'

screen shot 2016-10-10 at 13 10 31

views/index.hbs
```diff
@@ -26,70 +26,92 @@
...
+                  </div>
+                </div>
+              </a>
```
jenkoian: This is pretty hefty, can we put in a separate file and include it here? Similar to what we've done with error and prs below?

jenkoian:
@123aswin123 dunno, personally I'm not a fan of these kinds of buttons, but can understand some people like them for ease of use. I think if people want to share the main site they will just c/p the url. See some value in having custom intentions for personalised results though, so I'd just keep it on the results for now.

views/index.hbs
123aswin123: Okay Shall do....

123aswin123:
@jenkoian Gotcha! Please tell me if I can add the icons (along with personalized # of PRs ) in this position? http://i.imgur.com/JxRHzMw.png

jenkoian:
@123aswin123 yep perfect

123aswin123:
@jenkoian I've added personalized sharing for Twitter, to do personalized sharing on Facebook we need an App Key. Please let me know if it is necessary to do so.

jenkoian:
Thanks for your work on this @123aswin123 - I've decided to merge #54 mainly because they were first. However, if you would still like to contribute (and don't worry if not this will still count towards your hacktoberfest stats :)) then it'd be cool if you could add the twitter icon (and possibly the others too?) to the buttons added in #54. Once again thanks for your PR. 👍

PR #65, opened 2016-10-10
Minor wording changes

SemicolonExpected:
ln 3: "to see how close you are ..." sounds a little better than "for seeing how close you are"

ln 14: Stating "you can get involved" flows a little better than "get involved" which in this context seems like it doesn't fit as well. Also sounds more inviting.


jenkoian:
Cheers

PR #68, opened 2016-10-12
Frontend tests

matteo-hertel:
As discussed in #16 after the refactoring of the front-end JS I've added some tests using [Jest](https://facebook.github.io/jest/) 

The current code coverage is 75-80%

I've added NPM scripts as well:

test:  runs all the test suites (shortcut `npm t`)
test:watch: runs jest in watch mode,
test:coverage: runs jest with the code coverage flag,
test:updateSnapshot: runs jest and updates the stored snapshot,
test:full: runs jest in watch mode with the code coverage flag

once the back-end is refactored as well we can hook up jest to test all that as well

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,11 @@
   "description": "Small app to see if you have fulfilled the requirements for a Hacktoberfest t-shirt",
   "main": "index.js",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
+    "test": "node ./node_modules/jest/bin/jest ",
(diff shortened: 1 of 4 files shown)


jenkoian:
Thanks @matteo-hertel this looks really good, had not seen jest before but looks neat, particularly like the snapshot feature. Would be good to add something to the readme at some point (basically what you put in the PR description) but think it's fine for now. 👍

matteo-hertel:
I'll update readme with instructions and how to installato and use Jest, thanks for merging!

PR #70, opened 2016-10-12
add yarn.lock

matteo-hertel:
the big new since yesterday is the release of the new JavaScript dependency manager [Yarn](https://yarnpkg.com/), it works exactly like npm but it generates a yarn.lock to lock down the exact version of the dependencies, that file should be committed to VC so every developer will get the very same version of any dependency


jenkoian:
Well this is slightly terrifying, I guess jest brings in the kitchen sink huh! Thanks for the PR though, Yarn does indeed look sweet! Merging, and happy to leave the deps as they are for now, just something to keep an eye out for I guess 👍

PR #71, opened 2016-10-13
Live update

redice44:
This was fun. 👍 #23 

I didn't remove the /issues route even though it isn't being utilized anymore.

It should also sort by date created properly. It wasn't being resorted after it was added to the stack asynchronously.


jenkoian:
Thanks @redice44 this is amazing! Trying it locally and ran into an issue though. After so long (sometimes the second 'tick' sometimes more) the results seemed to duplicate. Are you able to replicate this also?

redice44:
Hmm, I'll take a look at it. I have an idea on what might produce that. When I was making it there were tons of updates happening, so it's quite possible I didn't notice any duplication. Thanks!

redice44:
It shouldn't be displaying duplicates anymore. 👍  I've also increased the polling time to 1 minute instead of 10 seconds. I'm not really sure what a good time would be, but it's easy to change

jenkoian:
@redice44 hey thanks dude, will try and have a look at this in more depth at some point today. 👍

jenkoian:
@redice44 this is looking really good - just letting it run locally for a bit to keep an eye on it, will report back 👍

index.js
```diff
@@ -95,6 +100,8 @@ function getPullRequests(username) {
 
 var totalIssues = 0;
 var octoberOpenIssues = [];
+var tempIssues = [];
+var prevQueryTime = '';
```
jenkoian: Doesn't seem to get used?

jenkoian:
Can you use 4 space tabs as per the rest of the app?

jenkoian:
Ok I've been watching it for a bit and it seems to work just fine. One thing, the numbers (Help other projects - Total: 7228) seem to go up more often than the list is updated. Any idea why? I know we rely on the count from the github API, but still seems a little strange? I guess it could be old issues that get re-opened or something?

redice44:
Yea, I think that number may be changed from older issues either adding, removing or as you said getting reopened.

jenkoian:
Cool, well this looks good to go! Thanks again, not sure when it will make it live because heroku and github seems to be having some issues but hopefully soon. 🍻

jenkoian:
@redice44 sorry had to revert as was having issues where user PRs count was being affected (e.g. you'd get results of 104 PRs etc.) I will take a look at this too, but are you able to pick up if you have a minute?

redice44:
I'll take a look at it too.

redice44:
Found it. 😓 Accidentally commented the array reset out at some point. Uncommented it and the PR count seems to be accurate now.

jenkoian:
@redice44 are you able to open a new PR?

redice44:
@jenkoian Sure, #79

PR #73, opened 2016-10-14
Trim username

drewvolz:
Removing spaces around the username prevents a duplicate search bar from appearing if the user enters a query of two spaces.


jenkoian:
Thanks @drewvolz!

PR #75, opened 2016-10-21
opens "Fork me on github" on new tab

sameer2800:
when we open a new page in new tab, there is more chance of visitors staying in the website . so , i changed target to blank which opens "fork me on github" on new tab .


jenkoian:
Thanks!

PR #80, opened 2016-10-24
Added a Heroku Button to the readme

cnHeider:


jenkoian:
Very nice, have seen this on other repos and always wanted to do this. Quick question, how does it work with the env vars? Do you have to login to heroku after you've deployed and add these?

cnHeider:
The app.json file is the implicit template of the repo.
https://devcenter.heroku.com/articles/heroku-button#using-an-implicit-template
In the app.json file of the PR, "env" key declares the key value pairs of env vars. https://devcenter.heroku.com/articles/app-json-schema#env
`"env": {
  "GITHUB_TOKEN": {
   "description": "Link: https://github.com/settings/tokens/new",
   "required": true
 }`
This can later be changed by logging in.

jenkoian:
Cool - I think currently the app doesn't work if no token is found, but we only use it for more favourable limit rates. So I think we can wrap the following auth section in a check to see if we have a token https://github.com/jenkoian/hacktoberfest-checker/blob/master/index.js#L44 this will mean that you could effectively deploy to Heroku with no changes required. You can then configured a token if you want an increased rate limit. Are you able to make this change in this PR? Alternatively I'll merge anyway and this can be looked at another time.

index.js
```diff
@@ -41,10 +41,14 @@ var github = new GitHubApi({
...
-    token: process.env.GITHUB_TOKEN
-});
+if(process.env.GITHUB_TOKEN){
+  github.authenticate({
+      type: "oauth",
+      token: process.env.GITHUB_TOKEN
+  });
+}else {
+  console.log('No GITHUB_TOKEN specified, do so to increase rate limit');
+}
```
jenkoian: Very minor thing, mind having consistent spacing around the if/else and 4 space tabs, e.g.

```javascript
if (process.env.GITHUB_TOKEN) {
    github.authenticate({
        type: "oauth",
        token: process.env.GITHUB_TOKEN
    });
} else {
    console.log('No GITHUB_TOKEN specified, do so to increase rate limit');
}
```

jenkoian:
Thanks @cnHeider !

PR #84, opened 2016-10-26
Updated readme

gezafisch:


jenkoian:
Thanks for the PR, think I prefer it how it was though 👍

PR #89, opened 2016-10-30
documentation fix!

SpEcHiDe:
the long query string is not readable by human beings


jenkoian:
See where you're coming from, don't think we need the additional +'s between the key/value though. So `created: ' + '...` could just be `created: ...` if that makes sense?

jenkoian:

PR #100, opened 2017-09-29
Style error messages to have white text

drewvolz:
Black on blue does not seem to be as friendly as white on blue. This PR tries to follow the style of text I've seen on the front page.

### DIFF
--- a/public/js/main.js
+++ b/public/js/main.js
@@ -100,7 +100,7 @@ HacktoberfestChecker.prototype.makeSpinner = function() {
  * create the necessary HTML to show an error message using fluent syntax
  */
 HacktoberfestChecker.prototype.makeError = function(error) {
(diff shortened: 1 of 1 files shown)


jenkoian:
Hi @drewvolz, thanks for this I'm wondering if we put the class on the h2 to match similar when it doesn't error, or possibly make all text white by default by setting the color on the body in the css or something. What do you think?

drewvolz:
@jenkoian I like your idea of making all body text white by default. I'll also try and take out all instances of `addClass('white')` and `class='white'` if you're okay with that.

jenkoian:
Closing this as the white text has been applied elsewhere, I still think it would be good to have this done across all text though, so feel free to open a new PR if you want. Cheers!

PR #102, opened 2017-10-01
update express server to use es6 functions such as object shorthand, …

butlerx:
…arrow functions and const

### DIFF
--- a/controllers/index.js
+++ b/controllers/index.js
@@ -1,3 +1,5 @@
+'use strict';
+
 /**
  * GET /
  */
--- a/controllers/index.js
+++ b/controllers/index.js
@@ -1,28 +1,22 @@
-'use strict';
-
 /**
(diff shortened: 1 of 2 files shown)


controllers/index.js
```diff
@@ -1,45 +1,37 @@
-'use strict';
```
jenkoian: I get the following error due to this removal `Block-scoped declarations (let, const, function, class) not yet supported outside strict mode` I think upgrading the version of nodejs will likely fix this, but I would like the app to support the LTS version of nodejs so would like to stick with the node v6 minimum. I think the PR can be merged, but could you just put back the `use strict`?

butlerx:
sorry im running latest node didn't think about lts support

jenkoian:
Thanks @butlerx !

PR #103, opened 2017-10-01
Mobile styling

elliotmassen:
While using the checker on my mobile I noticed that the text detailing my pull request was spilling out of the container. I've increased the width of the container when viewed on a smaller device.

| Before | After |
|:---:|:---:|
| Before | After |


jenkoian:
Thanks @elliotmassen !

PR #114, opened 2017-10-01
Added ESlint 

mog13:
added basic eslint

run with `npm run eslint`
fix basic issues with `npm run eslint-fix`

picked single quotes (hence the .js updates) but this can be changed in the .eslintrc.js

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,9 @@
     "node": "^6.11.1"
   },
   "scripts": {
-    "start": "node index.js"
+    "start": "node index.js",
+    "eslint": "eslint controllers/**/*.js public/**/*.js index.js",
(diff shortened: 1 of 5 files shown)


mog13:
conflict seem to only be over the quote changes. seems best to go with master and then re-run eslint-fix :)

package.json
```diff
@@ -26,7 +28,8 @@
     "lodash": "^4.17.4",
     "method-override": "^2.3.5",
     "morgan": "^1.7.0",
-    "q": "^1.5.0"
+    "q": "^1.5.0",
+    "eslint": "^4.8.0"
```
butlerx: should eslint not be a dev-dependency

mog13: yup my bad

PR #126, opened 2017-10-02
Removed '#' before PR Statement

SuyashLakhotia:
I guess this is more of a personal preference but I think the statements look better without the preceding '#'.


jenkoian:
Agreed - thanks @SuyashLakhotia !

PR #127, opened 2017-10-02
Modify tone of certain message statements

whitfin:
This PR will modify some of the statements to make their tone a little more positive; just a minor nitpick. I think a lot of people tend to have negative connotations with a period, so just tweaked that and tried to make it a little more motivational.

Feel free to merge if you're cool with the changes, or reject if you think they're unnecessary!

### DIFF
--- a/controllers/index.js
+++ b/controllers/index.js
@@ -60,7 +60,14 @@ exports.index = (req, res) => {
     findPrs(req.query.username).then(() => {
         let length = prs.length;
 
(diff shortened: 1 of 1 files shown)


jenkoian:
Sounds good to me @whitfin - what are your thoughts on adding some emoji to the output as well?

whitfin:
@jenkoian I personally don't like emoji but it's up to you :p

jenkoian:
😢  fair enough! 😆  Thanks for PR!

PR #132, opened 2017-10-02
Fixed style for empty user input

silence44:
Before message "Username cannot be blank." was black and now it's white and following the same style as other errors.


jenkoian:
Thanks @silence44 take a look at https://github.com/jenkoian/hacktoberfest-checker/pull/100 am wondering if we can handle this differently, by setting the default text as white.

jenkoian:
Going to merge this one as not yet any progress on #100 thanks @silence44 !

PR #133, opened 2017-10-02
Fixed typo in statements.

yuibun:
Removed an unnecessary space in the statement.

- It 's    ->    It's

PR #134, opened 2017-10-02
A wild space appeared

knyzorg:
There is an extra space. This PR removes it.

I work very hard.


jenkoian:
Covered by #133 but thanks for the PR @knyzorg !

PR #138, opened 2017-10-03
Tweak design

hosmelq:
This is the new look.

screen shot 2017-10-03 at 1 55 40 am

PR #143, opened 2017-10-03
Fixed #141 not reponsive on mobile

### LINKED ISSUE #141: not responsive on mobile
When I search my username account via smartphone, the PR results didn't show the exact same as web result, like this : 

issue1

ubatin:
- Fixed https://github.com/jenkoian/hacktoberfest-checker/issues/141

Adding on `CSS` at `./public/css/main.css`
```css
a.pr {
  display: -webkit-box;
  display: -ms-flexbox;
  display: flex;
  -webkit-box-pack: 'center';
      -ms-flex-pack: 'center';
          justify-content: 'center';
  -webkit-box-align: 'center';
      -ms-flex-align: 'center';
          align-items: 'center';
}
```

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -22,6 +22,18 @@ h1 {
   font-weight: normal;
 }
 
+a.pr {
+  display: -webkit-box;
+  display: -ms-flexbox;
+  display: flex;
+  -webkit-box-pack: 'center';
+      -ms-flex-pack: 'center';
+          justify-content: 'center';
+  -webkit-box-align: 'center';
+      -ms-flex-align: 'center';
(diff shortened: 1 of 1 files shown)


Metamist:
A problem with this (and many other) approaches for this is that if you scale down the window, the left column with the icon which is defined with `w-10` becomes too small for the column, thus the text and the icon will overlap. A better solution would be to have breakpoints for the icon and the text content to scale them accordingly.
Also, the content text should probably be left-aligned.

jenkoian:
I think #138 / #151 fixes this. Thanks for the PR though!

Issue #140, opened 2017-10-03
It should be better to show Github profile image for the person who haven't submitted any PR.


ankitjain28may:

Metamist:
Can be closed now, already solved! :+1:

jenkoian:
Thanks @Metamist

PR #144, opened 2017-10-03
User avatar is now shown for people without PRs

Metamist:
- https://github.com/jenkoian/hacktoberfest-checker/issues/140

This pull request will allow the web interface to show data about the user, which currently is only used for the avatar.
Previous method was by getting the avatar from the pull requests, but if none are found the user avatar would not be shown.
The `findPrs` function is now a bit more generalized by returning the prs and the user data as resolved promise parameters, rather than variables defined inside the request.

Some suggestions would be to generalize the `findPrs` even further by moving it out of the request (thus easing the load on the app each request) and possibly exporting it as well.

### DIFF
--- a/public/js/main.js
+++ b/public/js/main.js
@@ -101,7 +101,7 @@ HacktoberfestChecker.prototype.makeSpinner = function() {
  */
 HacktoberfestChecker.prototype.makeError = function(error) {
     return $('<div/>').append(
-        $('<h2/>', {
+        $('<h2 class="orange"/>', {
             text: error
         })
     );
--- a/controllers/index.js
+++ b/controllers/index.js
@@ -4,9 +4,6 @@
  * GET /
  */
 exports.index = (req, res) => {
-    const prs = [];
-    let userImage;
-
     function findPrs(username) {
         const _ = req.app.get('_');
         const q = req.app.get('q');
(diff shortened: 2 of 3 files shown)


public/js/main.js
```diff
@@ -101,7 +101,7 @@ HacktoberfestChecker.prototype.makeSpinner = function() {
  */
 HacktoberfestChecker.prototype.makeError = function(error) {
     return $('<div/>').append(
-        $('<h2/>', {
+        $('<h2 class="orange"/>', {
```
jenkoian: This has since been changed to be white, so can probably lose this?

jenkoian:
Thakns @Metamist ! Hopefully it won't affect rate limits (as will now be making twice the calls) so will keep an eye on it, but a great feature 👍

Metamist:
The github client is never authenticated with the github token provided in the `.env` file. It is using the public api, thus no rate limits! :+1:

public/js/main.js
Metamist: Did not mean to include this in this PR. Seems to have worked out anyway, my bad nevertheless. :)

jenkoian:
You are right! That seems like a bug though, the search api does appear to have rate limits? https://developer.github.com/v3/search/#rate-limit

jenkoian:
@Metamist fancy taking a quick glance at https://github.com/jenkoian/hacktoberfest-checker/pull/154 for me?

Metamist:
Actually seems like I was wrong. Fetching a user does also count towards the rate limit. Seems strange to me, as I read somewhere that using api that display data that is publicly visible on the website shouldn't count towards the rate limit.

PR #148, opened 2017-10-04
Add meta theme-color

petrkle:
Add theme color on android.
screenshot_20171004-063305


jenkoian:
Nice, thanks @petrkle !

PR #154, opened 2017-10-04
Add back github authentication.

jenkoian:
For better rate limits.


Metamist:
Looks good to me, but maybe we should consider authenticating as a github App rather than a personal access token?

jenkoian:
@Metamist probably, however I would prefer not to have users have to go through the rigmarole of oauth so quite like the simplicity of the current set up.

Issue #142, opened 2017-10-03
Checker's icons don't distinguish between Merged and Closed PRs


Zeustopher:
The checker's icons don't distinguish between Merged and Closed PRs. They both show as the Merged icon.

A Closed PR should use the "Request" icon but it should be colored Red. That is how Github represents them.

I _think_ the fix would be related to the file https://github.com/jenkoian/hacktoberfest-checker/blob/master/views/partials/prs.hbs. It appears that there are only two states defined - one for "open" and one for "closed".

See line `<div class="fl w-10 pa2 closed">`

There probably needs to be a third state defined so there is "open" (which is the green request icon), "merged" (which is the purple merge icon), and "closed" (which is a red request icon).

Happy Hacktoberfest!

Zeustopher:
Pictures are worth a thousand words:

![image](https://user-images.githubusercontent.com/11470532/31138441-b946c3c0-a82b-11e7-866c-d8421b67d044.png)

David-Byrne:
I can try give this one a shot :smile:

David-Byrne:
Not sure this is such an easy fix. The JSON returned from the API seems to have `state: 'closed'` set for both closed and merged pull requests, as can be seen in [this](https://api.github.com/repos/freedomotic/fd-user-manual/issues/9) and [this](https://api.github.com/repos/freedomotic/fd-user-manual/issues/15)...

We might be able to get the data with another API call but I don't know if it's worth the added complexity.

Zeustopher:
Drats - I thinks you're right... Too bad. Would have been a nice improvement.

4e554c4c:
I think this can be used to determine whether a pr is closed or merged: https://developer.github.com/v3/pulls/#get-if-a-pull-request-has-been-merged

PR #160, opened 2017-10-04
Adds Closed icon to differentiate Merged and Closed PRs -RE issue 142

David-Byrne:
https://github.com/jenkoian/hacktoberfest-checker/issues/142

Probably needs some work. Not sure if the benefit added by the icon justifies the increase in complexity...


jenkoian:
@David-Byrne this is looking good to me, let me know when you feel it's finished and I'll give it a more thorough review 👍

David-Byrne:
Thanks @jenkoian, I've fixed the linting errors so it's ok from that point of view.

It increases the number of requests to the Github API by a fair amount though but we can't seem to get around that. Up to you whether the benefit justifies this extra cost. 
I'd also have liked to have it cleaner than counting the number of promises that have been fulfilled, but the API just returns errors if a change hasn't been merged so that stops us from using Promise.all()

controllers/index.js
```diff
@@ -44,8 +44,41 @@ exports.index = (req, res) => {
 
                 prs.push(returnedEvent);
             });
+
+            const requests = [];
+            _.forEach(prs, pr => {
+                const [repoOwner, repoName] = pr.repo_name.split('/');
```
jenkoian: This errors for me locally, possibly due to the version of node I'm on, but I think we should support form node 6 and up. Are you able to change this?

David-Byrne: I thought destructuring was supported by version 6? See https://stackoverflow.com/questions/17379277/destructuring-in-node-js

I've no problem changing it though if needed

jenkoian:
@David-Byrne I want to merge this but want to add some logging of rate limits, so I can keep an eye on it, I won't get chance to do that today, so was wondering if you were able to add that? The response from github should include the amount of requests left, then it should just be a case of calling console.log.

controllers/index.js
```diff
@@ -21,6 +21,7 @@ exports.index = (req, res) => {
         ]).then(gitData => {
             const foundPrs = gitData[0];
             const user     = gitData[1];
+            console.log("API calls remaining: " + user.meta['x-ratelimit-remaining']);
```
jenkoian: Needs to be single quotes for the build :)

jenkoian:
Thank you @David-Byrne absolutely brilliant PR!

PR #161, opened 2017-10-04
Add small horizontal padding to footer

lex111:
For better display on small screen (iPhone 5 example).

| Before | After 
| - | - 
| image | image


jenkoian:
Thanks @lex111 !

PR #164, opened 2017-10-04
Refactor example env file

lex111:
1. `.env.example` is more familiar than `.env.dist`
2. Since this file is copied and an `.env` file is created from it, there are quotes and comments that are not needed, so the quotes are removed, and the comments are raised above the variables, otherwise they will be in the variable GITHUB_TOKEN.

### DIFF
--- a/.env.dist
+++ b/.env.dist
@@ -1,3 +1,0 @@
-GITHUB_TOKEN='YOUR_TOKEN' # https://github.com/settings/tokens/new
-NODE_ENV='production' # production or development
-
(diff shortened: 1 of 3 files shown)


jenkoian:
1. Do you have an example of this in the wild, .dist is the standard I'm used to.

2. Thanks!

lex111:
Well, it's clear from the title, right? What is `dist`? Well, in general, I personally saw myself using `example` more often. You can also see if you use the search:
- [.env.example](https://github.com/search?utf8=%E2%9C%93&q=%22.env.example%22&type=Code) - 283,959 results
- [.env.dist](https://github.com/search?utf8=%E2%9C%93&q=%22.env.dist%22&type=Code) - 28,034 results

jenkoian:
Thanks @lex111 !

PR #171, opened 2017-10-06
Add webfonts with subsetting

imagentleman:
Changes: 
This PR changes the webfont (Squealer Regular) used on the main title to a version with subsetting (the same original font but only including a subset of characters, in this case the ones in "Hacktoberfest Checker").

In order to do that:

1. Downloaded the ttf from the live website: https://hacktoberfestchecker.herokuapp.com/css/fonts/squealer-webfont.ttf
2. Used https://transfonter.org/ with the default settings, selected all formats and just included "Hacktoberfest Checker" in the character set.

Tested locally and seems to work the same as before (Chrome 61, Edge 40 and Firefox 56).

Benefits of the change (sizes before & after):

Before (original):
eot: 17kb 
svg: 60kb
ttf: 42kb
woff: 19kb
woff2: 14kb

After (with subsetting):
eot: 6kb 
svg: 5kb
ttf: 6kb
woff: 4kb
woff2: 3kb

I think the project is going to restrict its use of Squealer Regular to the characters in the main title, so this little optimization would be safe to implement.


jenkoian:
Anything to help those on small data plans is alright by me, thanks @imagentleman !

PR #172, opened 2017-10-06
Add word break to PR item

lex111:
Changes: Add word break to PR item

Note: In `tachyons`, this property is removed.

Screenshots for the change:

Before | After
| - | - 
image | image

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -31,3 +31,7 @@ h1 {
   flex-grow: 1;
   flex-shrink: 0;
 }
+
+.pr-item {
+  word-break: break-word;  
+}
--- a/views/partials/prs.hbs
+++ b/views/partials/prs.hbs
@@ -18,7 +18,7 @@
     {{#if prs.length}}
         <div class="br2 center mt3 shadow-1 overflow-hidden w-50-ns w-80-m w-90">
             {{#each prs}}
-                <div class="bg-white {{#if this.has_hacktoberfest_label}}hacktoberfest {{/if}}lh-copy pa3 flex bb b--gray">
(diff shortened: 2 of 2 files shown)


public/css/main.css
```diff
@@ -31,3 +31,7 @@ h1 {
   flex-grow: 1;
   flex-shrink: 0;
 }
+
+.pr-item {
+  word-break: break-word;  
+}
```
jenkoian: Doesn't seem to be documented in tachyons, but wonder if we can use `word-wrap`? https://github.com/tachyons-css/tachyons/blob/fd38efabc264542b1bc31f727e942736ea57ecd0/src/_word-break.css

lex111: @jenkoian They removed this, so this is not in the file https://unpkg.com/tachyons@4.8.0/css/tachyons.min.css, which is used.

jenkoian: hmm odd, wonder why they removed it? Either way this looks good then, thanks!

lex111: @jenkoian I do not know, but it's strange, I still have a other suggestion: let's include `tachyons` via `npm` (I mean add `tachyons` in `package.json`), and not as currently, using the remote site https://unpkg.com? How do you like that? This will be fine, for example, if this site is unavailable, and `tachyons` styles will not load. I will create PR if you agree.

jenkoian: I like the idea, I wanted to do this originally but couldn't work out how to get the path in the view, however I recently remembered we can use the path module for this. So yeah PR it would be good! :)

lex111: @jenkoian I wanted to do this using `app.use` in `index.js`, do you mean it?

jenkoian: Yeah exactly, combined with the path stuff, e.g. https://github.com/jenkoian/hacktoberfest-checker/blob/master/index.js#L75

PR #173, opened 2017-10-06
 Move form input down a little bit so the check button isn't overlapped

ldabiralai:
Changes: Pad username form down so that at 440px wide the check button isn't obscured by the "fork me on github" link

Screenshots for the change:

Before:

image

After:

image

### DIFF
--- a/views/index.hbs
+++ b/views/index.hbs
@@ -1,6 +1,6 @@
 ﻿<h1 class="f1 f-6-ns orange tc"><a class="orange no-underline" href="/">Hacktoberfest Checker</a></h1>
 
-<div class="tc">
+<div class="tc username-form">
     <form action="/" class="center h2 w-30-ns w-60-m w-80 flex" method="get">
         <input class="bn br--left br1 ph2 flex-auto" type="text" name="username" placeholder="GitHub username" value="{{ username }}">
         <button class="bg-orange bn br--right br1 pv0 ph3 pointer white" type="submit">Check</button>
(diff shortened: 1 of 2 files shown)


lex111:
You can solve this in another way: add `position: relative` to the parent.

views/index.hbs
```diff
@@ -1,6 +1,6 @@
 ﻿<h1 class="f1 f-6-ns orange tc"><a class="orange no-underline" href="/">Hacktoberfest Checker</a></h1>
 
-<div class="tc">
+<div class="tc username-form">
```
jenkoian: Can we use a tachyone class rather than define a new one? See http://tachyons.io/docs/layout/spacing/

ldabiralai:
> We can solve this in another way: add position: relative to the parent.

By doing this we can no longer click the "Fork on github" link, we'd then have to add `z-index` to the link which doesn't solve anything. Unless I'm misunderstanding what you mean?

views/index.hbs
ldabiralai: Ahh, sure, will update

lex111:
> By doing this we can no longer click the "Fork on github" link, we'd then have to add z-index to the link which doesn't solve anything. Unless I'm misunderstanding what you mean?

Can we of course, have you tried this option?

ldabiralai:
> Can we of course, have you tried this option?

I'm not sure what you mean, I tried with `relative` but that prevented the user from being able to click "Fork me on github" if they wanted to.

lex111:
@ldabiralai It's strange, in my Google Chrome, if I add `relative` class to the form container (`tc`), then the link from ribbon is not blocked.

jenkoian:
I get the same as @ldabiralai only a very small portion of the ribbon is clickable if I make the container relative. Any last comments before I merge?

lex111:
Well, then we leave it as it is now.

jenkoian:
Thanks @ldabiralai !

PR #176, opened 2017-10-06
Update description in link to generate GitHub token

lex111:
Changes: A tiny fix so that there is no confusion with Hacktoberfest, so there should be "Hacktober Checker".


jenkoian:
Good catch, thanks!

PR #177, opened 2017-10-06
Switch to Yarn

lex111:
Changes: Use `yarn` in README and Docker, since it is mainly used and remove lock files in `.gitignore`.


jenkoian:
Thanks @lex111 !

Issue #135, opened 2017-10-03
Add quick access to my personal account


Fleker:
If you add a login button and allow people to login through GitHub, it would allow the site to automatically display their current stats.

Or you could store the last-checked user as a cookie and automatically display that the next time the user opens the site.

jenkoian:
> it would allow the site to automatically display their current stats

Could you elaborate on this? Personally I find it must less effort to just chuck in my username than have to go through the two factor auth rigmarole.

Fleker:
When I go to hacktoberfestchecker.herokuapp.com it would redirect me to https://hacktoberfestchecker.herokuapp.com/?username=fleker

It'd be slightly more convenient, maybe not at the cost of the OAuth work.

You could just store the username in a cookie so when the user visits the main site or a certain page like https://hacktoberfestchecker.herokuapp.com/me it redirects them.

jenkoian:
@Fleker would be open to a PR for the cookie thing, as agree it could be a convenience :)

jenkoian:
Fixed in #183

PR #183, opened 2017-10-07
Adds a new /me URL that redirects to a username cached in LocalStorage

Fleker:
Fixes issue #135 by allowing users to save their username, allowing them to visit  https://hacktoberfestchecker.herokuapp.com/me to access faster.

Changes: 
* Adds `/me` path to `index.js`
* Creates a small `me.hbs` file to handle redirect
* Adds UI to the PR file which provides helpful context and a save button.

Screenshots for the change:
capture
captureme


ShavidzeT:
Great feature, but what about "Log In with GitHub" button? I thinks it's better and more flexible.

Fleker:
I had considered that, but this route seemed easier and wouldn't rely on setting up an OAuth flow. But I could change it if you think it would be preferable.

ShavidzeT:
We already have store app token in .env file so there is no problem from this side.

But agree, this route easier than github login integration.

If there are other users who needs this feature that would be great. 💯 

Let's see...

jenkoian:
Yeah I'm quite against github login because I think it's too heavyweight for the app which is pretty lightweight. I also like that you can check the score for your mates easily.

views/partials/prs.hbs
```diff
@@ -46,5 +46,18 @@
...
 {{/exists}}
 
+<script>
+    function storeUser() {
+        // Save username into localStorage. Recall username by visiting /me.
+        localStorage.myGitHub = document.querySelector('[name="username"').value;
+        // Provide some sort of visual feedback.
+        window.location.href='/me';
+    }
+</script>
```
jenkoian: How about moving this to `main.js`?

views/partials/prs.hbs
```diff
@@ -46,5 +46,18 @@
             {{/each}}
         </div>
     {{/if}}
+
+    <div class="bg-white br2 center mt5 shadow-1 overflow-hidden w-50-ns w-80-m w-90 mt2">
```
jenkoian: Nice one for using the tachyons classes! How about losing the white background here? I think it is secondary to the results and should be a bit more subtle in appearance.

views/me.hbs
```diff
@@ -0,0 +1,5 @@
+<script>
+// Get the cached user.
+const me = localStorage.myGitHub;
+window.location.href = `/?username=${me}`;
```
lex111: It's better to add a check for the existence of the GitHub login (to avoid `/?username=undefined`).

views/partials/prs.hbs
```diff
@@ -46,5 +46,18 @@
...
+        <button class="bg-orange white center mt2 h2 bn pointer br1 pv0 ph3 db-ns" onclick='storeUser()'>This is Me</button>
+        <p class='light-silver center tc-ns'>In the future, you can find your PRs by visiting <a href='/me' class='link underline-hover'>https://hacktoberfestchecker.herokuapp.com/me</a> on this device.</p>
+    </div>
 {{/exists}}
 
+<script>
+    function storeUser() {
+        // Save username into localStorage. Recall username by visiting /me.
+        localStorage.myGitHub = document.querySelector('[name="username"').value;
```
lex111: You can make it shorter by using `zepto`, which is already include:
```js
localStorage.myGitHub = $('[name="username"]').val();
```

views/partials/prs.hbs
Fleker: I was not sure if I should given the niche purpose of it.

views/partials/prs.hbs
Fleker: Should the color be something different, or just invert the text colors to show up on a dark background?

views/me.hbs
Fleker: Good catch. I'll fix it in the next revision.

views/partials/prs.hbs
Fleker: Thanks for that advice.

Fleker:
New style without the white box.

capture3

views/me.hbs
```diff
@@ -0,0 +1,5 @@
+<script src="/vendor/zepto/dist/zepto.min.js"></script>
+<script src="/js/main.js"></script>
+<script>
+redirectToUserPage();
```
jenkoian: Call this from `main.js` using a document ready type thing.

views/partials/prs.hbs
```diff
@@ -46,5 +46,9 @@
             {{/each}}
         </div>
     {{/if}}
-{{/exists}}
 
+    <div class="br2 center mt5 overflow-hidden w-50-ns w-80-m w-90 mt2">
+        <button class="bg-orange white center mt2 h2 bn pointer br1 pv0 ph3 db-ns" onclick='saveUserPage()'>This is Me</button>
```
jenkoian: Can we use an on click handler to do the save user paging, we have zepto included so should be able to do something like:

```js
button.on('click', function(e) {
    saveUserPage();
});
```

You may want to put an ID on the button.

views/me.hbs
Fleker: Should I check the URL and redirect if it matches?

jenkoian: Ah yeah that would be good to 👍

views/partials/prs.hbs
ShavidzeT: What about es6 arrow function feature?
```js
button.on('click', e => saveUserPage());
```

views/partials/prs.hbs
```diff
@@ -46,5 +46,9 @@
             {{/each}}
         </div>
     {{/if}}
-{{/exists}}
 
+    <div class="br2 center mt5 overflow-hidden w-50-ns w-80-m w-90 mt2">
+        <button class="bg-orange white center mt2 h2 bn pointer br1 pv0 ph3 db-ns" id="saveUser">This is Me</button>
+        <p class='light-silver center tc-ns'>In the future, you can find your PRs by visiting <a href='/me' class='link white underline-hover'>https://hacktoberfestchecker.herokuapp.com/me</a> on this device.</p>
```
jenkoian: Pass the host to the view (https://stackoverflow.com/questions/7507015/get-hostname-of-current-request-in-node-js-express) and use that rather than hard code production url (so it works the same on localhost)

Fleker: Right now the `href` goes to `/me` so it does work on localhost, although I can update the displayed text as well.

jenkoian:
Thanks @Fleker !

PR #186, opened 2017-10-08
Ignore pull request on repos owned by same user

printminion:
Fixes issue of users creating pull requests on own repos

Changes: Compare username with user name in pull request URL (case insensitive)

### DIFF
--- a/controllers/index.js
+++ b/controllers/index.js
@@ -30,6 +30,14 @@ exports.index = (req, res) => {
 
                 const hacktoberFestLabels = _.filter(event.labels, label => label.name.toLowerCase() === 'hacktoberfest');
 
+
+                //ignore pull requests on own repos
+                const repoOwner = repo.replace('https://github.com/', '').split('/')[0];
+
+                if (username.toUpperCase() === repoOwner.toUpperCase()) {
+                    return;
+                }
+
                 const returnedEvent = {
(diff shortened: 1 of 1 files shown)


lex111:
@printminion this is a good suggestion, it does **not** contradict the rules, and such PR will be taken into account. From Hacktoberfest support:

> You can contribute to your own repos. But, we encourage you to use Hacktoberfest as an opportunity to find a new project to contribute to.

But I think it is better able to visually highlight such PR's? Or to show separately from others?

jenkoian:
Agree with @lex111 some visual marker to show them as slightly less valued or something would be better than ignoring them completely.

controllers/index.js
```diff
@@ -30,6 +30,14 @@ exports.index = (req, res) => {
 
                 const hacktoberFestLabels = _.filter(event.labels, label => label.name.toLowerCase() === 'hacktoberfest');
 
+
+                //ignore pull requests on own repos
+                const repoOwner = repo.replace('https://github.com/', '').split('/')[0];
+
+                if (username.toUpperCase() === repoOwner.toUpperCase()) {
```
fobos: Every cycle in `_.each` you call `username.toUpperCase`. It's better to call it only once before `_.each`.

jenkoian:

PR #191, opened 2017-10-09
Disable spellcheck on username field

retroverse:
Disables spellcheck on username field to avoid the annoying (and ugly) squiggly red line. Tested on chrome.
screenshot 1


jenkoian:
Thanks @retroverse !

Issue #198, opened 2017-10-11
Add editorconfig support


jenkoian:
I favour 4 space tabs for pretty much everything. http://editorconfig.org/

PR #199, opened 2017-10-11
Add .editorconfig file

bqueue:
For issue #198

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,3 +5,5 @@ root = true
 [*]
 indent_style = space
 indent_size = 4
+end_of_line = lf
+insert_final_newline = true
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,7 @@
+# Top-most EditorConfig file
+root = true
+
+# 4-column tab indentation for every file
+[*]
+indent_style = tab
```
jenkoian: space please.

.editorconfig
```diff
@@ -1,7 +1,7 @@
 # Top-most EditorConfig file
 root = true
 
-# 4-column tab indentation for every file
+# 4-space indentation for every file
 [*]
```
jenkoian: Can we add the following as well?

```
end_of_line = lf
insert_final_newline = true
``` 
?

jenkoian:
Thanks @bqueue !

Issue #200, opened 2017-10-11
Hacktoberfest: To add Facebook share button to the website


Manveer695:
**Current behavior:**
No Facebook share button in the website

**Expected behavior:**
A Facebook share button or link should be available, when the score is shown.

Please add hacktoberfest tag to the issue. Thanks!

sharikovvladislav:
@jenkoian why  you added hacktoberfest tag? 

can I just add it to my projects too? :)

jenkoian:
@sharikovvladislav not sure I understand?

sharikovvladislav:
You added hacktoberfest label.  So now PR which will solve this issue will participate in hacktoberfest event.

Am I right? Only PRs which solves issue with that label will be counted for event?

jenkoian:
Any PR regardless of label or status count. The label is just suggested by Digital Ocean to help get people involved.

sharikovvladislav:
Ok thanks.

sinAshish:
close this.

biigpongsatorn:
https://github.com/jenkoian/hacktoberfest-checker/pull/210

Rabinzon:
@biigpongsatorn you have a bug:)

biigpongsatorn:
@Rabinzon  I fixed it.

jenkoian:
Fixed in #204

PR #204, opened 2017-10-11
feat: add facebook share button

Rabinzon:
Fixes issue #200 

Changes: Facebook share button added 

Screenshots for the change: http://take.ms/b8oQh  http://take.ms/eAmkj

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -39,10 +39,3 @@ h1 {
 .u-bgc--f4f4f4 {
   background-color: #f4f4f4;
 }
-
-.social-buttons {
-  display: flex;
-  width: 180px;
-  margin: auto;
-  justify-content: space-around;
-}
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -39,3 +39,10 @@ h1 {
 .u-bgc--f4f4f4 {
   background-color: #f4f4f4;
 }
+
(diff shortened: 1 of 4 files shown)


public/css/main.css
```diff
@@ -38,4 +38,11 @@ h1 {
...
-}
\ No newline at end of file
+}
+
+.social-buttons {
+  display: flex;
+  width: 180px;
+  margin: auto;
+  justify-content: space-around;
+}
```
jenkoian: Rather than add a new class, can we use tachyons utility classes? http://tachyons.io/

public/js/main.js
```diff
@@ -87,7 +87,7 @@ HacktoberfestChecker.prototype.getUsernameIssues = function(e) {
 HacktoberfestChecker.prototype.usernameIssuesSuccess = function(html, textStatus, xhr) {
     this.results.html(html);
     window.twttr.widgets.load();
-
+    window.FB.XFBML.parse();
```
jenkoian: Can we add a newline after. Alos may be nice to put this and the twitter one in a `loadSocialWidgets()` method or something?

Rabinzon: sure, one moment)

jenkoian:
Thanks @Rabinzon !

Issue #207, opened 2017-10-11
error in css: class .Stars-three


Thaynannadu:
**Current behavior:**
Stars animation does not work.
Error in css, class:**.stars-three**

**Steps to reproduce:**
![captura de tela 2017-10-11 as 13 57 40](https://user-images.githubusercontent.com/16005700/31454728-c73aa426-ae8c-11e7-8470-98cb732067b9.png)

jenkoian:
See #213 and #214

PR #208, opened 2017-10-11
little changes in css

Thaynannadu:
Fixes issue #error in css: class .Stars-three #207

remove "top: 20% left: 20%" in .stars-three. to work properly. I think.

add padding-top in title for mobile.

before:
captura de tela 2017-10-11 as 14 23 41

after:
captura de tela 2017-10-11 as 14 23 29

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -34,3 +34,7 @@ h1 {
 .u-bgc--f4f4f4 {
   background-color: #f4f4f4;
 }
+
+.ph2-ns {
+  padding: 20vh 0 0 0;
+}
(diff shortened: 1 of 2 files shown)


public/css/main.css
```diff
@@ -33,4 +33,8 @@ h1 {
 
 .u-bgc--f4f4f4 {
   background-color: #f4f4f4;
+}
+
+.ph2-ns {
+  padding: 20vh 0 0 0;
```
jenkoian: I don't think we should use the same class name. If we can't use some other padding class (https://github.com/tachyons-css/tachyons-spacing) then we should add a custom class.

jenkoian:

PR #209, opened 2017-10-11
Feature: Use modified version of the hackoberfest logo

kalebo:
I have modified the header so that it more closely matches Hacktoberfest's official logo. 

The svg data has been optimized and is embedded directly in index.hbs. I have left the old css for the squealer font in place just in case it is wanted elsewhere.

2017-10-11 12_42_28-hacktoberfest checker

Note that this has some side-effects that are similar to #208; alias the fork me banner will not cover the logo when the window is resized.


jenkoian:
Thanks for this @kalebo. I quite like the logo of this tool being slightly different to the main logo though and thus would prefer to keep them distinct.

Issue #213, opened 2017-10-12
The page layout doesn't look good because of the stars and clouds style


kenma9123:
**Current behavior:**
With the newly added style UI for the stars and cloud the page layout doesn't look good. Because of the scrollbars. 
![image](https://user-images.githubusercontent.com/1641990/31484029-2f0d6c58-af61-11e7-8aa4-8c9a6fa3a2ce.png)

**Expected behavior:**
it should be a perfect page with no scroll bar since we are only dealing a single section.

**Steps to reproduce:**
Just open the page and try to drag one of the scrollbars

jenkoian:
Hopefully fixed by #214 can someone confirm?

f-breidenstein:
Tested in FF and Chrome. Looks good.

PR #214, opened 2017-10-12
3rd star background should blend on the section size

kenma9123:
Fixes issue #213

Changes: Because of the `top` and `left` style prop of the `.stars-three` class it extends page layout size that looks like on the issue. Instead of removing the said prop. I added a `width` and `height` to compensate the excess size. Thus work on several screen size.

Screenshots for the change:
image

### DIFF
--- a/public/css/stars.css
+++ b/public/css/stars.css
@@ -27,6 +27,8 @@
 /*Had to add more stars and offset them*/
 
 .stars-three {
+    width: auto !important;
+    height: auto !important;
     top: 20%;
     left: 20%;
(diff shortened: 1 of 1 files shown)


jenkoian:
What browser/OS you using, as I don't seem to get this. Thanks for the fix though, will merge 👍

kenma9123:
chrome/windows

jenkoian:
Think possibly a windows issue with the scrollbars? Anyway pushed your fix up, let me know if it's now working, cheers for the PR!

kenma9123:
yup it's fixed now.

PR #217, opened 2017-10-12
Cookie added for username

SoumyadeepJana:
Fixes issue #135

Changes: [Added cookie for the username so that a user need not type in his username everytime he/she visits the app.]

Screenshots for the change:


jenkoian:
See #183

PR #219, opened 2017-10-12
Update README.md

chardon1:
Add a link to Hactoctoberfest official site

Fixes issue #[Add issue number here. If you do not solve the issue entirely, please change the message e.g. "First steps for issues #IssueNumber]

Changes: [Add here what changes were made in this issue and if possible provide links.]

Screenshots for the change:

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,9 @@ As an alternative to the section above, you can run the app within a Docker cont
 
 * `$ docker run -p 5000:5000 -e "GITHUB_TOKEN=YOUR_TOKEN" hacktoberfest-checker`
 
+## Link
+Check your progress on [Hacktoberfest official site](https://hacktoberfest.digitalocean.com/stats/)
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,6 +41,9 @@ As an alternative to the section above, you can run the app within a Docker cont
 
 * `$ docker run -p 5000:5000 -e "GITHUB_TOKEN=YOUR_TOKEN" hacktoberfest-checker`
 
+## Link
+Check your progress on [Hacktoberfest official site](https://hacktoberfest.digitalocean.com/stats/)
```
jenkoian: I'm not sure about this, it might be confusing having a link to an alternative checker in the README. If we are going to add it, we should explain better what it is and how it is different. What do you think?

jenkoian:
Closing, favour having it on the app such as in #245

PR #224, opened 2017-10-13
Fixed documentation typos

daphnied:
Fixes issue #[194]. 

Changes: [Fixed CONTRIBUTING.md typos.]

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,13 +1,13 @@
 # Contribute
 
 ### Step 1
--  Fork the repository.
+-  Fork the repository
 
 ### Step 2
 - Create a feature branch (git checkout -b your-new-feature)
 
 ### Step 3
-- Commit the changes (git commit -m 'Add some feature') (will add the new changes)
+- Commit the changes (git commit -m 'Add some feature') (will additionally add the new changes)
 
 ### Step 4
 - Push to the branch (git push origin my-new-feature)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,17 +1,16 @@
 # Contribute
 
 ### Step 1
--  Fork the repository
+-  Fork the repository.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,17 +1,16 @@
 # Contribute
 
 ### Step 1
--  Fork the repository
+-  Fork the repository.
```
jenkoian: If we're going to do this, should probably do it for all steps, personally don't think we need the full stop.

CONTRIBUTING.md
```diff
@@ -1,17 +1,16 @@
...
 
 ### Step 2
 - Create a feature branch (git checkout -b your-new-feature)
 
 ### Step 3
-- Commit the changes (git commit -am 'Add some feature') (will additional add the new changes)
+- Commit the changes (git commit -m 'Add some feature') (will add the new changes)
```
jenkoian: I think the old one is fine, but additional should be additionally.

CONTRIBUTING.md
```diff
@@ -1,17 +1,16 @@
...
 
 ### Step 4
 - Push to the branch (git push origin my-new-feature)
 
 ### Step 5
 - Create pull request
-
```
jenkoian: Can we keep the newline

daphnied: Yes that is fine

CONTRIBUTING.md
```diff
@@ -7,11 +7,10 @@
 - Create a feature branch (git checkout -b your-new-feature)
 
 ### Step 3
-- Commit the changes (git commit -am 'Add some feature') (will additional add the new changes)
+- Commit the changes (git commit -m 'Add some feature') (will additionally add the new changes)
```
jenkoian: Need the `a` too otherwise changes won't be added (as the bit in brackets suggests).

PR #228, opened 2017-10-14
Updated Views

Manukam:
Changes: [Added "Status" to the title.]

Screenshots for the change:


jenkoian:
Would rather keep it as is sorry.

PR #232, opened 2017-10-15
Change username input color

romulomourao:
Changes: Add `orange` and `f4` classes to input field

Screenshots for the change:
selecao_311


jenkoian:
Thanks but I'm not sure about the orange text sorry.

PR #234, opened 2017-10-16
Error on non-User user types

HarrisonGrodin:
This PR enforces the fact that organizations cannot submit pull requests or participate in Hacktoberfest. Previously, searching for an organization would simply display that it had completed 0/4 PR's.

### Changes
 - Displays error message when a non-`User` user `type` is received from the GitHub API (e.g. `Organization`).

### Screenshots
Previous behavior:
Previous behavior

New behavior:
New behavior


controllers/index.js
```diff
@@ -23,6 +23,11 @@ exports.index = (req, res) => {
             const user = gitData[1];
             const prs = [];
 
+            if (user.data.type !== 'User') {
```
jenkoian: For an organisation `user.data.type` won't exist (I think?) so do we need to check for that as well?

HarrisonGrodin: Through my testing, it seems that `user.data.type` will be `'Organization'`.

controllers/index.js
```diff
@@ -23,6 +23,11 @@ exports.index = (req, res) => {
...
+                deferred.reject('not user');
```
jenkoian: Not that it really matters, but can we use `notUser` just find whitespace can sometimes make things a bit confusing.

HarrisonGrodin: Done.

controllers/index.js
```diff
@@ -140,7 +145,9 @@ exports.index = (req, res) => {
         }
     }).catch((err) => {
         console.log(err);
-        if (req.xhr) {
+        if (err === 'notUser') {
+            res.status(400).render('partials/error-user', {layout: false});
```
jenkoian: What happens if you hit refresh having put in an organization name? Does it display a white page with just the error in black? If so, I think this needs to be within the `if (req.xhr)` conditional.

jenkoian:
Thanks @HarrisonGrodin are you able to resolve the conflict? Will merge then 👍

HarrisonGrodin:
Conflicts resolved.

controllers/index.js
```diff
@@ -51,9 +56,15 @@ exports.index = (req, res) => {
...
-                res.status(404).render('partials/error', {layout: false});
+                if (err === 'notUser') {
+                    res.status(400).render('partials/error-user', {layout: false});
+                } else {
+                    res.status(404).render('partials/error', {layout: false});
+                }
             } else {
-                res.render('index', {error: true, username});
+                res.render('index', {
+                  error: true, errorUser: err === 'notUser', username
```
jenkoian: How about always passing the message in error rather than boolean then we can just use the existing error stuff and don't need all the conditions around errorUser and an additional partial?

HarrisonGrodin: Good call. Done. :+1:

jenkoian:
Thanks @HarrisonGrodin !

Issue #235, opened 2017-10-16
"This is me" functionalist broken in WebKit (Safari)


josh-:
**Current behavior:**
- Clicking "This is me" throws the following exception in WebKit-based browser:

```
[Error] SyntaxError (DOM Exception 12): The string did not match the expected pattern.
	querySelectorAll (zepto.min.js:2:4807)
	qsa (zepto.min.js:2:4807)
	init (zepto.min.js:2:4103)
	r (zepto.min.js:2:4359)
	saveUserPage (main.js:162)
	(anonymous function) (main.js:169)
	proxy (zepto.min.js:2:16155)
```

Furthermore, since this exception is thrown, the value isn't set and the page then doesn't redirect to `/me`

**Expected behavior:**
- Clicking on the "This is me" button saves the current username and redirects to me `/me`

**Steps to reproduce:**
1. Open the site in a WebKit-based browser
2. Click "This is me"
3. Look at browser console

jenkoian:
Fixed in #236

PR #236, opened 2017-10-16
Fix broken selector in Webkit-based browsers

josh-:
Fixes issue #235

Changes: Add closing brace to `name="username"` selector, so that WebKit no longer throws an exception ([the only browser to do so](https://www.reddit.com/r/firefox/comments/5nbmqi/on_handling_invalid_selector_strings:)).

Screenshots for the change: N/A


jenkoian:
Thanks @josh- !!

PR #245, opened 2017-10-18
Added a notice about the official checker to the footer

aaronosher:
Changes: Added a notice to the footer about the official Hacktoberfest checker.
Screenshot: 
screen shot 2017-10-18 at 5 44 21 pm

### DIFF
--- a/views/partials/footer.hbs
+++ b/views/partials/footer.hbs
@@ -1,4 +1,3 @@
 <footer class="f6 ph2 tc flex-none">
-    <p class="white">Disclaimer: This site is fan made and not affiliated with <a class="white" href="https://hacktoberfest.digitalocean.com/" target="_blank">Hacktoberfest</a>.</p>
-    <p class="white">Note: There is now an official <a class="white" href="https://hacktoberfest.digitalocean.com/" target="_blank">Hacktoberfest</a> checker. It can be found <a class="white" href="https://hacktoberfest.digitalocean.com/stats/" target="_blank">here</a>.</p>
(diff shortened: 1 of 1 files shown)


jenkoian:
I'm in two minds about this one. It's kind of saying 'oh hey don't use this checker, use this other one instead' but on the other hand, I'm not trying to compete with Digital Ocean and in fact have been in touch with them about this since before Hacktoberfest started. 

I think I'm going to leave this PR open for a bit and see if there are any strong opinions either way, then make a final decision later on. Thanks for the PR though!

aaronosher:
I actually agree. While I generally prefer the aesthetic of the official one, I really like how this one actually shows the status of the pull request, which the Digital Ocean one doesn't. 

But I put it at the bottom, opposed to next to the search bar, just because I think it makes more sense as a kind "heads up", but it's not telling the user to go there, just advising.

One benefit of the official one, is that it only shows your stats if you're signed up, and seeing as that's internal to Digital Ocean, it's somewhat important.

jenkoian:
> One benefit of the official one, is that it only shows your stats if you're signed up, and seeing as that's internal to Digital Ocean, it's somewhat important.

Yeah good point. Ok, happy to go with this then, would you mind changing slightly though so it's part of the disclaimer, e.g.

```html
Disclaimer: This site is fan made and not affiliated with <a class="white" href="https://hacktoberfest.digitalocean.com/" target="_blank">Hacktoberfest</a>. Check <a class="white" href="https://hacktoberfest.digitalocean.com/stats" target="_blank">here</a> for your official statistics
```

What would be really neat too is if we only show the second  part if a username has been submitted, we can then link to their distinct stats page.

aaronosher:
Edited, and that should be reasonably simple to do. Digital Ocean gives an HTTP/1.1 404 status code if the account isn't registered.

jenkoian:
@aaronosher sorry I don't mean submitted to digital ocean, I just mean submitted through the form on this web app...

aaronosher:
@jenkoian I'm aware. It'd just be an async request to DO. Show link if status code 200. Otherwise, Show "not signed up" if status code 404.

jenkoian:

PR #249, opened 2017-10-20
use eslint to set code spacing

butlerx:
Changes: I noticed that there was some inconsistent spacing with tabs and some spaces being 2 or 4 so i set eslint to do consistent spacing

### DIFF
--- a/public/js/.eslintrc.js
+++ b/public/js/.eslintrc.js
@@ -1,0 +1,12 @@
+module.exports = {
+    env: {
+        browser: true,
+        commonjs: true,
+        es6: false,
+        jquery: true,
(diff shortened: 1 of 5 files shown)


public/js/.eslintrc.js
```diff
@@ -0,0 +1,12 @@
...
+    browser: true,
+    commonjs: true,
+    es6: false,
+    jquery: true,
+    node: false,
+  },
+  rules: {
+    'no-console': ['warn'],
+  },
+};
```
jenkoian: This should be 4 space indentation

jenkoian:
Thanks @butlerx !

PR #252, opened 2017-10-23
Also save user when clicking on /me link

dwaq:
Changes: After searching for a username, there is an orange button "This is me", that when clicked, saves the username into local storage. I added the saveUser click handler to the http://hacktoberfestchecker.herokuapp.com/me so clicking on that link also saves the username to local storage.

I added this because I clicked on the http://hacktoberfestchecker.herokuapp.com/me link before I clicked on the "This is me" button and it didn't work (although I expected it to). I don't think this change will cause any negative reactions because just typing and going to http://hacktoberfestchecker.herokuapp.com/me doesn't have an affect.

Screenshots for the change: No visual changes

### DIFF
--- a/views/partials/prs.hbs
+++ b/views/partials/prs.hbs
@@ -75,6 +75,6 @@
 
     <div class="br2 center mt5 overflow-hidden w-50-ns w-80-m w-90 mt2">
         <button class="bg-orange white center mt2 h2 bn pointer br1 pv0 ph3 db-ns" id="saveUser">This is Me</button>
-        <p class='light-silver center tc-ns'>In the future, you can find your PRs by visiting <a href='{{ hostname }}/me' class='link white underline-hover'>{{ hostname }}/me</a> on this device.</p>
+        <p class='light-silver center tc-ns'>In the future, you can find your PRs by visiting <a href='{{ hostname }}/me' class='link white underline-hover' id="saveUser">{{ hostname }}/me</a> on this device.</p>
     </div>
 {{/exists}}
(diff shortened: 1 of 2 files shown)


views/partials/prs.hbs
```diff
@@ -75,6 +75,6 @@
 
     <div class="br2 center mt5 overflow-hidden w-50-ns w-80-m w-90 mt2">
         <button class="bg-orange white center mt2 h2 bn pointer br1 pv0 ph3 db-ns" id="saveUser">This is Me</button>
-        <p class='light-silver center tc-ns'>In the future, you can find your PRs by visiting <a href='{{ hostname }}/me' class='link white underline-hover'>{{ hostname }}/me</a> on this device.</p>
+        <p class='light-silver center tc-ns'>In the future, you can find your PRs by visiting <a href='{{ hostname }}/me' class='link white underline-hover' id="saveUser">{{ hostname }}/me</a> on this device.</p>
```
jenkoian: As it's bad practice to re-use ids I think we should change this, perhaps we can put the id on the surrounding div and then target the button and the a tag from that or something? What do you think?

dwaq: I didn't think of that. Good point. Adding the ID to the surrounding div makes the entire "In the future,..." text clickable, which absolutely works but is a weird UX. Would it be OK to change saveUser into a class and make the click handler work on that class? I've tested that this works, but I'm unsure if this is also bad practice.

jenkoian: Was thinking you could just target the button and the a, e.g. `$('#saveUser').find('button, a')` or something (note: pseudo code). But yeah, use class if easier 👍

dwaq: The latest commits use the class method I described.

jenkoian:
Thanks @dwaq !

Issue #254, opened 2017-10-24
the PR limit


divyam3897:
**Current behavior:**

If you have PR greater than 30 it just shows the latest 30/4

**Expected behavior:**
It should show all the Pull requests.

jenkoian:
@divyam3897 you've been too busy! 😆  I guess this is because we're only hitting the first page of results from the API. Will put a hacktoberfest label against this if anyone wants to pick it up, else I'll take a look if I get chance. Thanks for reporting!

rvben:
I just made a PR to fix this issue.

krider2010:
Haha oops - I was working on one at the same time!

krider2010:
My change supports paging through all results rather than just pushing the `per_page` value, if the maintainer wants to offer that?

jenkoian:
Both excellent PRs so merged them both, thanks! :)

PR #260, opened 2017-10-25
Page through search data

krider2010:
Fixes issue #254.

Changes: Changed the promise chain slightly, due to the annoying way the API works. So there are a few extra `Promise.resolve` calls in the original code as well as the obvious new code.

Screenshots for the change:
screen shot 2017-10-25 at 21 45 36


jenkoian:
Thanks @krider2010 !

Issue #248, opened 2017-10-19
Uncaught SyntaxError: Unexpected token ILLEGAL main.js:74: backtick (`) illegal in "old" browser


Modified:
**Current behavior:**

As the title says: "Uncaught SyntaxError: Unexpected token ILLEGAL main.js:74".
**Expected behavior:**

Transpile it for JS (ES5), duh?
**Steps to reproduce:**

Run (ie open this site/app) in an "older" browser.

jenkoian:
What "older" browser are you using?

jenkoian:
Fixed in #262

PR #262, opened 2017-10-26
Babel transpilation for front-end files.

davcs86:
Fixes issue #248 

Changes: Adds babel transpilation before of start task


jenkoian:
Thanks @davcs86 !

PR #269, opened 2017-10-31
Check usernames with @ symbol

GEveloper:
When I was checking friends progress with copy/paste-ing their usernames, experienced this issue to check them with @ symbol. Below commit will fix the problem.

### DIFF
--- a/controllers/index.js
+++ b/controllers/index.js
@@ -26,11 +26,9 @@ const errorCodes = {
  */
 exports.index = (req, res) => {
     const github = req.app.get('github');
-    const username = req.query.username;
-
-    if (username.charAt(0) == '@') {
-        username = username.substring(1);
-    }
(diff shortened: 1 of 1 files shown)


controllers/index.js
```diff
@@ -28,6 +28,10 @@ exports.index = (req, res) => {
     const github = req.app.get('github');
     const username = req.query.username;
```
jenkoian: As you're re-assigning need to make this let.

jenkoian:
Thanks @GEveloper I get an internal server error on the main page (without searching a username) - the error is:

```
TypeError: Cannot read property 'length' of undefined
    at exports.index (/Users/ian/Code/hacktoberfest-checker/controllers/index.js:29:40)
```

Are you able to fix?

jenkoian:

PR #270, opened 2017-10-31
Fix 'This is Me' button alignment on small devices

gabrielecker:


public/css/main.css
```diff
@@ -50,3 +50,6 @@ h1 {
     box-shadow: inset 0px 0px 0px 2px #ff6300;
 }
 
+.saveUser {
+    display: block;
+}
```
jenkoian: Please use a utility class, see https://github.com/tachyons-css/tachyons-display so simple add class `db` on to the element.

jenkoian:
Thanks @gabrielecker !

PR #274, opened 2018-09-30
2018 edition

giovannipessiva:
Here is an update for Hacktoberfest 2018 edition 🍻

### DIFF
--- a/controllers/index.js
+++ b/controllers/index.js
@@ -7,7 +7,8 @@ const moment = require('moment');
 const statements = [
     'It\'s not too late to start!',
     'Off to a great start, keep going!',
-    'Half way there, keep it up!',
+    'Keep it up!',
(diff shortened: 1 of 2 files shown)


andys8:
Build is failing. Seems to be setup related.

jenkoian:
Hey thanks for this, I'm currently in the process of rewriting the app for 2018. I am however way behind on that, so will possibly merge this as an interim fix. Just one thing though, are you able to change the messages as for the 2018 edition you need 5 PRs as opposed to 4. Thanks!

jenkoian:
Cheers for heads up @giovannipessiva that's pretty cool! Think you can only view people's progress if they've registered, which makes sense, but sometimes people may wanna check their progress to see if it's even worth registering, or whatever, so think this still has a place!

PR #276, opened 2018-10-01
replace hardcoded year, show last year's result before October

tessus:
- use the current year, show last year's result before October 
- make background animation optional

This change removes the hardcoded year to search for PRs.
It also shows last year's result when checking before October.

The animation of the background (stars and clouds) is using a lot of CPU on the client machine.
The entire system becomes rather sluggish. This is not a great experience.

Animation is now off by default.
To activate it again, set the ENV variable ANIMATION to yes in the file .env

### DIFF
--- a/controllers/index.js
+++ b/controllers/index.js
@@ -3,7 +3,6 @@
 const _ = require('lodash');
 const moment = require('moment');
 
-const year = 2018;
 const prAmount = 5;
 
 const statements = [
@@ -24,6 +23,7 @@ const errorCodes = {
     notUser: 400
 };
 
+
 /**
  * GET /
  */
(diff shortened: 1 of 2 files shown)


tessus:
give me just a sec. I saw that `getNextPage` uses deprecated stuff. let me fix this first.

tessus:
Now all is good ;-)

jenkoian:
Hey thanks for this @tessus, I've actually just got rid of the starts, so if you rebase you can probably remove the animation check. Really like the current year check though, so happy to merge once you've taken out the animation stuff :)

tessus:
Here you go!

jenkoian:
Great stuff, thanks @tessus !

Issue #281, opened 2018-10-01
Display number of PRs opened within hacktoberfest


jenkoian:
If you look at https://hacktoberfest.digitalocean.com/details it mentions the first 50,000 registered people will be eligible for the t-shirt. Now obviously we don't have access to the number of people registered but it may still be quite nice to display the total number of PRs opened in October. This is of course if we can indeed get such information from the Github API.

chgasparoto:
I'll take a look.

omergulen:
It seems like we can't do direct search on GitHub API, hence we need to find issues created after October 2018 and labeled with hacktoberfest and then find their PR's.

tanmayrajani:
> It seems like we can't do direct search on GitHub API, hence we need to find issues created after October 2018 and labeled with hacktoberfest and then find their PR's.

Wouldn't that be misleading? I mean, the issue doesn't need to be opened in October (AFAIK), it doesn't need to contain the Hacktoberfest label.

omergulen:
AFAIK it needs to be, also we need to get PR's, not issues as in #292. It needs to be fixed, correct me if I am wrong.

mtpjr88:
is someone working on this or can I take it?

chgasparoto:
@mtpjr88 there is a discussion on #292

sidvishnoi:
In case no one has noticed it, Hacktoberfest site has an API:
https://hacktoberfest.digitalocean.com/api/statistics

jenkoian:
@sidvishnoi good spot! :) suspect using that would be against their T&C's but worth asking I think.

sidvishnoi:
Looking at that `last_updated_at` timestamp, looks like they update it once per day. Maybe we can store and update it by a cron job (or even manually)

PR #292, opened 2018-10-01
Adds a counter for all PRs opened for Hacktoberfest 2018

chgasparoto:
Fixes issue #281 - It only appers in the screen after hit the search button.

Changes: Created a new request to search for issues with the params: type: pr, label: hacktoberfest, created: > 2018-09-30, created: < 2018-11-01. Extract data from the field `total_count`.

Screenshots for the change:
selection_027

### DIFF
--- a/controllers/index.js
+++ b/controllers/index.js
@@ -54,16 +54,18 @@ exports.index = (req, res) => {
 
     Promise.all([
         findPrs(github, username),
+        loadPrsOpened(github),
         github.users.getForUser({username})
             .then(logCallsRemaining)
     ])
-        .then(([prs, user]) => {
+        .then(([prs, openedPrs, user]) => {
             if (user.data.type !== 'User') {
                 return Promise.reject('notUser');
             }
 
             const data = {
                 prs,
+                openedPrs: openedPrs.prs_already_opened,
                 isNotComplete: prs.length < prAmount,
                 statement: getStatement(prs),
                 username,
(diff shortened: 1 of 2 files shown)


controllers/index.js
```diff
@@ -150,6 +152,23 @@ function loadPrs(github, username) {
     return promise;
 }
 
+function loadPrsOpened(github) {
+    return new Promise((resolve, reject) => {
+        github.search.issues({
+            q: 'type:pr+label:hacktoberfest+created:>2018-09-30+created:<2018-11-01'
```
jenkoian: *Any* PR opened in October counts, so don't think we need the label here. Also, can you use the times as well, as done here https://github.com/jenkoian/hacktoberfest-checker/blob/master/controllers/index.js#L130 just to keep in consistent.

views/partials/prs.hbs
```diff
@@ -1,3 +1,9 @@
+{{#exists openedPrs}}
+<div class="f6 ph2 tc flex-none">
+    <p class="white">PRs already open for 2018: <strong>{{ openedPrs }}</strong>.</p>
```
jenkoian: Can you change the wording slightly... `PRs opened so far this month:` Also can we put the openedPrs through a number format, you can either do this in the node (probably easiest) or use a handlebars helper like https://gist.github.com/DennyLoko/61882bc72176ca74a0f2.

controllers/index.js
tanmayrajani: @jenkoian Also, it's still slightly misleading in a way that not all the users that have created PRs have signed up for Hacktoberfest as well. Only if we can find that out as well... :smiley:

jenkoian: Yeah, I wrote about that in [the issue](https://github.com/jenkoian/hacktoberfest-checker/issues/281) we can't really get that info, so just a useful barometer more than number of participants, maybe we should add some wording to explain?

chgasparoto: I'll fix them.

And I agree in giving some explanation. Any ideias?

views/partials/prs.hbs
chgasparoto: Sure :+1:

controllers/index.js
chgasparoto: I don't think this approach will work. I'm getting a lot of timeouts (504) because the number of PRs increase since yesterday. I'll seach another alternative for it.

chgasparoto: The issue isn't the rate limit but the time to process the request. It's return error 504.

jenkoian: Hmm, not sure why that would happen. However, if you tweak the query to take out the hacktoberfest label, you get millions upon millions, so perhaps not that useful. It may be useful if we can do PRs opened by unique users, but not sure how easy/feasible that would be.

PR #294, opened 2018-10-02
Better Docker configuration 

thecreazy:
Changes: 

- **Added the .dockerignore:** when you in the Dockerfile do the `COPY ./APP` command it will copy all the file, also the .git folder, that are useless and so much big 

- **Added the docker-compose:** for a local usage or also to not use a docker command I had also add the docker-compose configuration

- **Fix the Dockerfile:** the application need node > 8.x but in the dockerfile is used the node 6-alpine

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,9 +3,7 @@ version: "3"
 services:
   hacktoberfest:
     build: .
-    environment:
-      GITHUB_TOKEN: YOUR_TOKEN
-      NODE_ENV: production
-      GA_CODE: YOUR_CODE
+    env_file:
+      - .env
     ports:
       - 5000:5000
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,11 @@
+version: "3"
+
+services:
+  hacktoberfest:
+    build: .
+    environment:
+      GITHUB_TOKEN: YOUR_TOKEN
+      NODE_ENV: production
+      GA_CODE: YOUR_CODE
```
jenkoian: Could you load in the env file direct? https://docs.docker.com/compose/environment-variables/#the-env_file-configuration-option

thecreazy: Done, added the use of the .env

jenkoian:
Nice, thanks!

PR #301, opened 2018-10-03
added 404 route to handle 404 errors

joshterrill:
Changes: I added a catch all route to handle 404's gracefully.

Screenshots for the change:

### DIFF
--- a/controllers/index.js
+++ b/controllers/index.js
@@ -209,3 +209,7 @@ const logCallsRemaining = res => {
 exports.me = (req, res) => {
     res.render('me');
 };
+
+exports.notfound = (req, res) => {
+    res.render('404');
+};
--- a/index.js
+++ b/index.js
@@ -78,6 +78,7 @@ app.use('/vendor', express.static(path.join(__dirname, './node_modules')));
 
 app.get('/', IndexController.index);
 app.get('/me', IndexController.me);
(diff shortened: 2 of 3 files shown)


joshterrill:
Yeah I just used the same font styles as other places in the app, but maybe it doesn't work here. You want me to change them to both use the serif font?

joshterrill:
@jenkoian  Sorry, do you mean the header alone looks like it's mixing fonts? I can see what you mean, but all I did was apply the same style that's on the other page. It appears that the font that was chosen does not have a shape for every character in the alphabet... look at my screenshot below where I type "ABCDEFG" 





So, what I can do for now is just remove that "fancy" font and go with all serif font since that font doesn't have all the letters of the alphabet in it.

joshterrill:
@jenkoian This has been fixed in my latest commit. Here's a screenshot of what it looks like now:



That is super weird that the other `squealerregular` does not have every letter of the alphabet in it... was it just by chance that all the letters in "Hacktoberfest" were supported? Haha!

jenkoian:
Yeah, not sure about the font, it's an open font so surprised it doesn't support all letters?! 

Can you provide a link back to the homepage from the 404, happy to merge then.

joshterrill:
@jenkoian Done.

PR #303, opened 2018-10-03
fix: Change color for the .orange CSS class for A11y purpose (contrast)

Gillespie59:
Fixes issue #[Add issue number here. If you do not solve the issue entirely, please change the message e.g. "First steps for issues #IssueNumber]

Changes: [Add here what changes were made in this issue and if possible provide links.]

Screenshots for the change:

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -60,3 +60,7 @@ h1 {
 .bn.br--left.br1.ph2:focus {
     outline: none;
 }
+
+.orange {
+    color: #a52607;
+}
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -61,6 +61,6 @@ h1 {
     outline: none;
 }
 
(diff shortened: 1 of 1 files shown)


jenkoian:
This seems to make contrast worse on the dark main body bit, did you just want this applied when on a white background?

public/css/main.css
```diff
@@ -60,3 +60,7 @@ h1 {
 .bn.br--left.br1.ph2:focus {
     outline: none;
 }
+
+.bg-white .orange {
+    color: #a52607;
+}
```
jenkoian: Thanks! The banner at the top is still affected as it uses bg-white and orange on the same div. So maybe needs to be done on `.bg-white.orange` as well.

jenkoian:
Thanks!

PR #299, opened 2018-10-02
converted from tachyons to tailwind

zjael:
Fixes issue #286

Changes: 
Converted from tachyons to tailwind - as close as i could
Added npm script for building tailwind css file

Screenshots for the change:
Original:
original
Tailwind version:
tailwind

PR #305, opened 2018-10-03
Fix invisible avatar

crhallberg:
`.o-0 glow` makes the avatar invisible to start. I've replaced it with `.dim` to make the avatar visible (and match the screenshots) while still having a reactive component.

This will also be fixed by #299 ([line number](https://github.com/jenkoian/hacktoberfest-checker/pull/299/files#diff-506a5a0ceb78b91144bb23a742b4c7f8R25)).

Before:
image

After:
image


jenkoian:
Nice patch, thanks!

Issue #304, opened 2018-10-03
Update Pull Request template


jenkoian:
I don't think the existing pull request template is great, so would like to update it. Am open to ideas of what it should be, or even if we need one at all?

One PR template I came across recently which I really liked was this: https://raw.githubusercontent.com/timber/timber/master/.github/PULL_REQUEST_TEMPLATE.md although the messages will likely need to be tweaked as this project isn't as big as that one.

deneuv34:
Cool!
> One PR template I came across recently which I really liked was this: https://raw.githubusercontent.com/timber/timber/master/.github/PULL_REQUEST_TEMPLATE.md although the messages will likely need to be tweaked as this project isn't as big as that one.

About the sample template. I want to discuss, how about we add the `Screenshot` section for optional section?

yeshijinsel:
I wanna take this issue

jenkoian:
@yeshijinsel about to close this one as someone has already picked it up, thanks though!

PR #165, opened 2017-10-04
Add issue template and improve pull request template

divyam3897:
Added an issue and pull request template in .github/ following https://github.com/blog/2111-issue-and-pull-request-templates

PR #306, opened 2018-10-03
Updating pull PULL_REQUEST_TEMPLATE.md from issue #165

deneuv34:
Fixes issue #304 

Changes: I just updating Pull Request Template file into format that references on the issue.

Screenshots for the change: 
screenshot from 2018-10-04 00-17-21

If there is message still need to change, please let me know

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,10 +6,6 @@ First off, hello! Thanks for submitting a PR. We love/welcome PRs (especially if
 <!-- Description of the problem that this code change is solving -->
 
 
-#### Solution
-<!-- Description of the solution that this code changes are introducing to the application. -->
-
-
 #### Usage Changes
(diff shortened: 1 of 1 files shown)


jenkoian:
Hmm the more I think about this the more I think maybe we just ditch it. The project is too small to require so much information and it could actually put people off, WDYT?

deneuv34:
@jenkoian yeah, how about remove `Consideration` & `Solution` section? :thinking:

jenkoian:
@deneuv34 thanks, happy to give that a go, just one last thing, do you think you can add a sentence reminding people to add themselves to the `humans.txt` file?

deneuv34:
@jenkoian I don't know. I think it'll better if we give reminder on the contribution guideline, mostly people will read it before make PR. That's what I though, what do you think?

jenkoian:
Yeah alright, will merge, thanks!

deneuv34:
you're welcome

PR #310, opened 2018-10-03
reverse proxy support

### LINKED ISSUE #277: doesn't run behind a reverse proxy
**Current behavior:**
Several files are not found. e.g.: `tachyons.min.css`, `main.css`, ...

**Expected behavior:**
It should show the web site correctly (the same way as if it were run locally).

**Steps to reproduce:**

- Start with `yarn start`
- Add the following to your local MAMP/LAMP installation (or your production web server):
    ```
    ProxyPass "/hacktoberfest" "http://localhost:5000"
    ProxyPassReverse "/hacktoberfest" "http://localhost:5000"
   ```
- browse to http://localhost/hacktoberfest

tessus:
This change involves several parts:

- replace absolute paths with relative ones
- use the referer to figure out the hostname, if behind a reverse proxy
- if no referer is available in a reverse proxy setup, use another workaround to create the proper /me link

It also fixes the following error `TypeError: res.sendStatus is not a function`.

Fixes #277


controllers/index.js
```diff
@@ -48,6 +47,27 @@ exports.index = (req, res) => {
...
     }
 
+    // in a reverse proxy situation we have to use the referer to retrieve
+    // the correct protocol, hostname, and path
+    // unfortunately this won't work, when accessng the page directly:
+    // e.g.: http://example.com/hacktoberfest/?username=XXX
+    // in such a case we set hostname to an empty string and create the link
+    // with js after the page has loaded
+    if (req.headers['x-forwarded-for']) {
+        //console.log(JSON.stringify(req.headers));
```
jenkoian: Can you remove this leftover debug?

controllers/index.js
```diff
@@ -48,6 +47,27 @@ exports.index = (req, res) => {
...
+        const referer = req.headers.referer;
+        if (referer) {
+            var hostname = referer.split("?")[0].slice(0, -1);
+            if (hostname.endsWith('/m')) {
+                hostname = hostname.slice(0, -2);
+            }
```
jenkoian: Can you just explain what' going on here? Will this still work if not hosted on a .com?

controllers/index.js
tessus: sure, my bad. this was not on purpose.

views/index.hbs
```diff
@@ -1,8 +1,8 @@
-﻿<h1 class="f1 f-6-ns tc"><a class="white no-underline" href="/">Hacktoberfest Checker</a></h1>
+﻿<h1 class="f1 f-6-ns tc pb4"><a class="white no-underline" href="">Hacktoberfest Checker</a></h1>
 
 <div class="tc pb4">
     <p class="white">{{ timemessage }}</p>
-    <form action="/" class="center h2 w-30-ns w-60-m w-80 flex" method="get">
+    <form action="" class="center h2 w-30-ns w-60-m w-80 flex" method="get">
```
jenkoian: Is this needed? I know it probably just submits to the same page, but looks strange with a blank action.

tessus: This I have to test. I did it because I knew it would work this way.

controllers/index.js
tessus: Yes, it will still work on a .com domain.

This is a rather interesting code path:

First it checks the referer and creates the hostname from it. This would be the flow when clicking the `Check` button. (Usually the hostname would also have the trailing slash, but since you already used the variable hostname without the slash, I also removed the trailing slash.)

However, if you click on the `.../me`link or browse directly to the `.../me` link, the referer will be the `.../me` link (due to the redirectin your app). In the step above I removed the last character, because I assumed it's the trailing slash. In case of the `.../me` link that's not the case, thus I have to check for `/m` instead of for `/me` and then remove the last 2 characters to get the hostname.

Then the last possibility is, if a user directly accesses the page `https://example.com/subdir/?user=XXX`, in which case there won't be a referer at all.
I create an empty hostname, so that I can apply the last workaround after the page has loaded.

This entire code path is only used when in a reverse proxy setup.

I could have written it differently, but it would have had more lines and more tests/checks. I'm a performance guy. The less instructions, the better.

views/index.hbs
tessus: Yes, it seems to work, even though it seems wrong. I will add it back though, since it's your code.

tessus:
I made the changes you have requested.

controllers/index.js
jenkoian: I guess not a big issue as only done with the x-forwarded-for set but this seems a little flaky to me, i.e. if we wanted to change the /me url to be something like /yourstats for example, this would break? I don't have any better solutions as I've not really experienced the issues you have behind a reverse proxy.

tessus: Yes, in that case it would break. But if you wanted to change `/me` to `/yourstats`, you would have to change the `/me` code in several places, so I don't think this one line would make a difference. 

Also, please try to run it behind a reverse proxy and not in the root of the virtual host. Then your entire app will break. (e.g. `http://localhost/subdir/` or `https://hacktoberfestchecker.herokuapp.com/checker/`) 

It's clear that you haven't experienced these issues, because you are running it in the root of  a virtual host. But you can't expect that from all of your users.

tessus: I'm running it [here](https://evermeet.cx/hfc/) and it works perfectly. I also tested it without reverse proxy and when running in the root of a virtual host. It doesn't change anything for your setup. It only makes it location agnostic thus you can now run your app on:

`https://example.com`
`https://example.com/hacktoberfest/`
`https://example.com/apps/hfc/`
and so on.

Without my change, your app can only be run as `https://example.com`.

jenkoian: Good responses :) off to bed now but will take a closer look in the morning, in the mean time are you able to fix the build?

tessus: eslint makes a few errors. this is not my fault. But I can explain this tomorrow, if you want.

Btw, I will make a small change, so that future changes to `/me` won't break the code. You only set a var and that's it.

tessus:
I fixed the positive eslint errors, even though the code is now a bit less readable:

```
var link = window.location.href.split("?")[0] + 'me';
```
is definitely more readable than
```
var link = `${window.location.href.split('?')[0]}me`;
```

There are still 2 false positives and one that is not an error:

```
/home/travis/build/jenkoian/hacktoberfest-checker/controllers/index.js
  65:17  error  'hostname' is already defined  no-redeclare
  68:13  error  'hostname' is already defined  no-redeclare
/home/travis/build/jenkoian/hacktoberfest-checker/index.js
  85:29  error  'next' is defined but never used  no-unused-vars
```

**On line 65 and 68** `hostname` is not re-declared but declared for the first time. These if blocks are mutually exclusive thus 3 separate code paths. eslint made an error. These are the false positives.

**Line 85**: yes, it's true that next is not used, but I had to add it, otherwise the arguments are interpreted incorrectly, which leads to the error `TypeError: res.sendStatus is not a function`.
Also, I am not going to use a var, just to get rid of this error. What am I supposed to do? Print the var to `/dev/null`? I think not.

Here an excerpt from the express API doc:

_Error-handling middleware always takes four arguments. You must provide four arguments to identify it as an error-handling middleware function. Even if you don’t need to use the next object, you must specify it to maintain the signature. Otherwise, the next object will be interpreted as regular middleware and will fail to handle errors. For details about error-handling middleware._

controllers/index.js
```diff
@@ -48,6 +48,26 @@ exports.index = (req, res) => {
...
+    // the correct protocol, hostname, and path
+    // unfortunately this won't work, when accessng the page directly:
+    // e.g.: http://example.com/hacktoberfest/?username=XXX
+    // in such a case we set hostname to an empty string and create the link
+    // with js after the page has loaded
+    if (req.headers['x-forwarded-for']) {
+        const referer = req.headers.referer;
+        if (referer) {
+            var hostname = referer.split('?')[0].slice(0, -1);
+            if (hostname.endsWith(statsLink.slice(0, -1))) {
```
jenkoian: Can we rename hostname here, so `url` or something? Then we can leave the hostname definition above which you removed and we can lose an else.

tessus: Yep, I see what you mean. Will do in a bit. Gonna grab lumch first. ;-)

tessus:
Ok, all set.

tessus:
Btw, we could get rid of hostname completely. As far as I could ascertain there's no real use for it anyway, except for creating the `/me` link and text. This could be entirely done in the page's js.

I just didn't want to change your code too much. In any case, if you consider this in the future, let me know and I'll create a separate PR.

tessus:
> It is used for some of the meta tags,

I see, thanks.

> Agree about the eslint next var thing, can you configure eslint to ignore this rule? 

Yes, but I will only change it to ignore the check for arguments. In other cases it makes sense to get rid of unused variables.

tessus:
Done.

public/js/main.js
```diff
@@ -151,24 +151,31 @@ function redirectToUserPage() {
...
     // Save is bound to button press
     $(document).on('click', '.saveUser', () => saveUserPage());
     // Works with /me or /me/
-    if (window.location.pathname.startsWith('/me')) {
+    if (window.location.pathname.endsWith('/me')) {
         redirectToUserPage();
     }
+    // if hostname is empty, use location to build the link
+    // needed for reverse proxy setup
+    if (document.getElementById('melink').innerHTML == '/me') {
```
jenkoian: This causes console errors for me on the main page where there is no me link, are you able to fix? Also, we have jquery (well zepto) available so you should be able to just do `$('#meLink')`

tessus: Sorry for that. I missed it.

>  Also, we have jquery (well zepto) available so you should be able to just do $('#meLink')

I have no idea what that means. This is the first time I've done something with js...

jenkoian:
Sorry it took so long, but all tested and we're good, thanks for the patch!

tessus:
You are welcome. Logging patch will be there in a bit.

Issue #309, opened 2018-10-03
Sort out npm scripts.


jenkoian:
There are a few npm scripts, some of which do things like compile with babel etc. However, it seems we're using this in dev then committing the compiled assets. I think we should probably be ignoring these from git and doing this as part of the start task.

I'm wondering if anyone fancies picking this housekeeping up?

mstruebing:
I'll take care :+1:

mstruebing:
@jenkoian you need to close this issue manually.

jenkoian:
Fixed in #311

PR #311, opened 2018-10-03
Remove compiled assets from repo

mstruebing:
Fixes issue #309.

Changes: I removed the compiled assets from the repository as they do not need to be versioned.

Everything seems to work fine with this change, I can start the server see the page and grab my username.


jenkoian:
Thanks! Presumably you had to run the start scripts before you ran yarn start (or node index.js)? If so, I'm wondering if you could update the start script to run these as part of that too? Deployment happens automatically on deploy to master and it just runs yarn start (actually it runs node index.js but I'm thinking maybe we update this https://github.com/jenkoian/hacktoberfest-checker/blob/master/Procfile).

mstruebing:
yarn start is working without compiling before because of the `prestart` script which is automatically executed by npm before start (similar to post scripts) you can read more about it here: https://docs.npmjs.com/misc/scripts.

Also `node index.js` is working for me without having these compiled assets.

I've made updated the PR to use `yarn start` inside it's `procfile` because it's more convenient I think.

jenkoian:
@mstruebing thanks! I did not know that! (not a node guy, despite this being my most popular repo which is indeed a node app hehe) that makes sense though. All in all, top patch, many thanks!

Issue #312, opened 2018-10-03
Use yarn in travis instead of npm


mstruebing:
I've seen that you are using npm in travis but you are providing a `yarn.lock` - which indicates that you are primarly using yarn and there is even no `package-lock.json` file.

So - travis should be using what devs and production is using to give reliable results.

It would be nice if you would add the hacktoberfest label to these issue :D

jenkoian:
Agreed and done 👍

PR #313, opened 2018-10-03
Use yarn in travis instead of npm

mstruebing:
Fixes issue #312 

Changes: I've exchanged the usage of npm inside the `.travis.yml` so that travis is using yarn instead of npm.


jenkoian:
Thanks!

PR #316, opened 2018-10-04
Copy files into image as user octocat

FuriKuri:
#### Issue
Files which will be copied to the `/app` directory in the docker image, are owned by `root`. They should be owned by the `octocat`, who will be used to start the server.

#### Usage Changes
There are no usage changes.

#### Screenshot
No visible changes.


jenkoian:
Thanks!

PR #320, opened 2018-10-07
Add fade-in to pr items

retroverse:
This is just adding a slight fade-in to the main body and pr-items using css3 animations. Not sure if this is 100% the right approach but might be a change worth considering.

Cheers.


jenkoian:
@retroverse are you able to provide a gif (using something like https://www.cockos.com/licecap/) of the change?

jenkoian:
Have checked it out and possible just too subtle, but I don't see the animation.

PR #322, opened 2018-10-08
- fix .gitignore

### LINKED ISSUE #318: Excessive logging in production
**Current behavior:**
When `NODE_ENV` is set to `production`, all requests are still logged to stdout. This is rather excessive.
I also don't really care how many PR there have been found and how many API calls are remaining.

**Expected behavior:**
This information should only be printed for a non production environments or when an env var ist set.

**Steps to reproduce:**
run with `NODE_ENV=production`

There are a few ways to fix this. The 2 most promising options are either the module `debug` or `winston`. 
If you let me know which one you prefer, I could have a PR ready in no time (Friday night).

tessus:
no comments are allowed after patterns (comments are only valid on their own line)

- remove excessive logging for 'production'

fixes #318

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 # Node stuff
 npm-debug.log
 node_modules
-package-lock.json # Favour yarn.lock, heroku fails if both included
+package-lock.json
 
 # API keys
 .env
--- a/index.js
+++ b/index.js
@@ -70,8 +70,17 @@ app.locals = {
     }
 };
 
+// Production error handler
+if (process.env.NODE_ENV == 'production') {
+    app.use((err, req, res, next) => {
+        console.error(err.stack);
+        res.sendStatus(err.status || 500);
+    });
+} else {
+    app.use(logger('dev'));
+}
+
 app.use(compression());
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -1,7 +1,7 @@
 # Node stuff
 npm-debug.log
 node_modules
-package-lock.json # Favour yarn.lock, heroku fails if both included
+package-lock.json
```
jenkoian: I think the comment is still worth keeping, can we just move it to the line above?

tessus: yes, that will work. I just thought it's not necessary. I'll add it back.

index.js
```diff
@@ -70,8 +70,17 @@ app.locals = {
...
 
+// Production error handler
+if (process.env.NODE_ENV == 'production') {
+    app.use((err, req, res, next) => {
+        console.error(err.stack);
+        res.sendStatus(err.status || 500);
+    });
+} else {
+    app.use(logger('dev'));
+}
```
jenkoian: How come moved up here?

PR #326, opened 2018-10-09
remove hardcoded hostname (link) for share buttons

tessus:
The twitter and fb share buttons use a hardcoded link.
This change uses the actual hostname (and path) of the checker web site.

### DIFF
--- a/public/js/main.js
+++ b/public/js/main.js
@@ -177,5 +177,7 @@ $(document).on('ready', () => {
         var link = `${window.location.href.split('?')[0]}me`;
         document.getElementById('melink').innerHTML = link;
         document.getElementById('melink').href = link;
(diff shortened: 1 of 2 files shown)


tessus:
Unfortunately the buttons don't work anymore since commit 50e350ec17aeb759916f3099b295d8387597c588

I rebased, since you changed master since then with commit https://github.com/jenkoian/hacktoberfest-checker/commit/50e350ec17aeb759916f3099b295d8387597c588. Then I noticed that the share buttons go to the wrong links.

Something must have broken the sharing functionality. Maybe the twitter and fb js is now invalid?

tessus:
btw, this has nothing to do with my change.  if you apply my change to 62537219e5cd35a20105647f8bbf573b8fc0471a it works perfectly.

jenkoian:
@tessus thanks for looking, working on this now 👍

PR #328, opened 2018-10-13
Add 'autocapitalize' and 'autocorrect' attributes to the input

tomodwyer:
#### Issue
On an iOS device the mobile keyboard autocapitalises the first letter of your username and also tries to autocorrect any words. This is not ideal for GitHub user names.

This fix removes that and shows a lower case keyboard with no autocorrect.

##### Before
before

##### After
after

#### References
- [Touch Keyboard Types](https://baymard.com/labs/touch-keyboard-types)
- [MDN HTML <form> element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attr-autocapitalize)


jenkoian:
Fantastic PR thank you!

PR #329, opened 2018-10-14
Store hostname as an env var rather than try to work out dynamically

jenkoian:


app.json
```diff
@@ -1,17 +1,30 @@
...
+    "logo": "https://github.com/jenkoian/hacktoberfest-checker/blob/master/public/img/og.png?raw=true",
+    "success_url": "/",
+    "env": {
+        "GITHUB_TOKEN": {
+            "description": "Link: https://github.com/settings/tokens/new",
+            "required": true
+        },
+        "APP_URL": {
+            "description": "URL of the site where this is deployed to",
+            "required": true
```
tessus: Why is this required? Shouldn't it be optional?

.env.example
```diff
@@ -6,3 +6,6 @@ NODE_ENV=production
 
 # code used to hook up google analytics
 GA_CODE=YOUR_CODE
+
+# url of the current environment
```
tessus: It should be mentioned somehow that this is only important for a reverse proxy setup. And maybe give an example.

tessus:
Other than the 2 comments above, it looks great. Thanks.

PR #330, opened 2018-10-16
Trim name input before storing in localstorage

zack6849:
You can have a space in the name, and the page will work fine and load, but if you hit "this is me" it will save it with a space and then break.


jenkoian:
Thanks.

Issue #319, opened 2018-10-05
React Version of Hacktoberfest Checker


frextrite:
**Current behavior:**

Couldn't find the branch with the react version as stated in the README

**Expected behavior:**

expected to find a branch with react version of the website as stated in the README

**Steps to reproduce:**

check branches

Maybe open source the react version so that the community could contribute to it.

jenkoian:
yep, sorry been really short of time, the react branch is really early/small but will get that up soon promise!

frextrite:
Great! Eagerly waiting for it.

AhsanAyaz:
@jenkoian I'd love to do it if the title had `Angular` 😄 

_All frameworks are good_

jenkoian:
Merged! Thanks again @rafaelklaessen

rafaelklaessen:
:smile:

PR #333, opened 2018-10-20
Rewrite in React + Refactor

rafaelklaessen:
This is a refactor of the hacktoberfest-checker in React. The backend is now a very light API, separate from the frontend, which is a React app (managed by create-react-app).

The next step would be to figure out what the best way to run the frontend is...Maybe served by the backend, or ran on a different server. Shouldn't be difficult.

Run the frontend with `yarn start` and the backend with `yarn server`.

I look forward to your reply! I know this certainly isn't done but would love to eventually see it up and running on your server.

As per screenshot...Everything still looks exactly the same :smile:


rafaelklaessen:
Note how the _public folder is the unused one

jenkoian:
Reviewing this now @rafaelklaessen first thoughts... WOW

> The next step would be to figure out what the best way to run the frontend is...Maybe served by the backend, or ran on a different server. Shouldn't be difficult.

Outside of simply `yarn server && yarn start` something like https://www.npmjs.com/package/parallelshell ?

api/controllers/pr/index.js
```diff
@@ -0,0 +1,51 @@
+'use strict';
+
+const _ = require('lodash');
```
jenkoian: This doesn't appears to be used so can be removed.

rafaelklaessen:
That would be fine for development, but I meant the actual deployment. Originally you had 1 server, now you'll need 2 (static file server for React and a server for the backend).
Don't know if that's a problem for you :smile:

api/controllers/pr/index.js
```diff
@@ -0,0 +1,51 @@
...
+const { getStatusCode, getErrorDescription } = require('./errors');
+
+/**
+ * GET /
+ */
+exports.index = (req, res) => {
+  const github = req.app.get('github');
+  const username = req.query.username;
+
+  var hostname = process.env.APP_URL || `${req.protocol}://${req.headers.host}`;
```
jenkoian: Not used can be removed.

package.json
```diff
@@ -21,27 +22,30 @@
...
     "tailwindcss": "^0.6.6"
   },
-  "repository": "git@github.com:jenkoian/hacktoberfest-checker.git"
+  "repository": "git@github.com:jenkoian/hacktoberfest-checker.git",
+  "browserslist": [
+    ">0.2%",
+    "not dead",
+    "not ie <= 11",
+    "not op_mini all"
+  ]
```
jenkoian: I didn't know about browserslist before, nice!

public/index.html
```diff
@@ -8,50 +8,25 @@
     <meta name="description" content="Useful app to check how you're getting on in hacktoberfest">
     <meta property="og:title" content="Hacktoberfest Checker">
     <meta property="og:description" content="Useful app to check how you're getting on in hacktoberfest">
-    <link rel="canonical" href="{{ hostname }}">
-    <meta property="og:url" content="{{ hostname }}">
+    <link rel="canonical" href="%REACT_APP_HOSTNAME%">
+    <meta property="og:url" content="%REACT_APP_HOSTNAME%">
     <meta property="og:site_name" content="Hacktoberfest Checker">
-    <meta property="og:image" content="{{ hostname }}/img/og.png">
+    <meta property="og:image" content="%PUBLIC_URL%/img/og.png">
```
jenkoian: This is something provided by react scripts right? Does it differ from `REACT_APP_HOSTNAME`? Should we just use `REACT_APP_HOSTNAME`?

src/components/PageWrapper.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const PageWrapper = ({ children, ...props }) => (
```
jenkoian: As we're not using babel (and don't particularly want to at this stage) and the spread operator isn't support in IE11 (which is included in our target browser list) I don't think we can use this without some sort of shim. WDYT?

src/components/UsernameForm/index.js
```diff
@@ -0,0 +1,61 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import withRouter from 'react-router-dom/withRouter';
+import TimeMessage from './TimeMessage';
+import UsernameInput from './UsernameInput';
+import CheckButton from './CheckButton';
+
+class UsernameForm extends Component {
+  static propTypes = {
```
jenkoian: This comes up an ESLint error for me, see https://stackoverflow.com/a/43426331/380054 this is similar to the spread operator thing I mentioned above. We have a few options to fix, do you have any preference?

src/pages/User/components/PullRequests/MeLinkInfo.js
```diff
@@ -0,0 +1,34 @@
...
+      <button
+        className="bg-teal-lighter text-pink-darkest mx-auto mt-2 h-8 border-none pointer rounded-sm px-4 block saveUser"
+        onClick={this.storeUsernameAsMe}
+      >
+        This is Me
+      </button>
+      <p className="text-grey-dark mx-auto text-center my-4">
+        In the future, you can find your PRs by visiting{' '}
+        <a
+          href="/me"
```
jenkoian: Should we prefix with the hostname?

src/pages/User/components/PullRequests/MeLinkInfo.js
```diff
@@ -0,0 +1,34 @@
...
+          className="link text-orange underline-hover saveUser"
+          id="melink"
+        >
+          {window.location.origin}
```
jenkoian: Can we use the hostname?

src/pages/User/components/PullRequests/ShareButtons.js
```diff
@@ -0,0 +1,44 @@
...
+import PropTypes from 'prop-types';
+import pullRequestAmount from './pullRequestAmount';
+
+const ShareButtons = ({ username, pullRequestCount }) => (
+  <div className="pb-8 flex justify-center">
+    <div className="p-2" id="twitter-share">
+      <a
+        target="_blank"
+        className="bg-blue text-white rounded px-2 py-1 pointer text-white no-underline text-sm"
+        href={`https://twitter.com/intent/tweet?text=My progress on hacktoberfest ${pullRequestCount} / ${pullRequestAmount}&url=${window.location.origin}/username/${username}&hashtags=hacktoberfest, hacktoberfestchecker`}
```
jenkoian: Shall we use the hostname in place of `window.location.origin`?

src/pages/User/components/PullRequests/ShareButtons.js
```diff
@@ -0,0 +1,44 @@
...
+      >
+        <i className="fab fa-twitter fa-lg" />
+        {' '}
+        Tweet
+      </a>
+    </div>
+    <div
+      className="p-2"
+      id="fb-share"
+      data-href={`${window.location.origin}/username/${username}`}
```
jenkoian: Shall we use the hostname in place of `window.location.origin`?

src/pages/User/components/PullRequests/ShareButtons.js
```diff
@@ -0,0 +1,44 @@
...
+      data-layout="button"
+      data-size="large"
+    >
+      <a
+        target="_blank"
+        className="fb-xfbml-parse-ignore bg-blue-dark text-white rounded px-2 py-1 pointer text-white no-underline text-sm"
+        href={`https://www.facebook.com/sharer/sharer.php?u=${window.location.origin}/username/${username}`}
```
jenkoian: Shall we use the hostname in place of `window.location.origin`?

jenkoian:
I pushed some spacing fixes which ESLint as well as my IDE were reporting. 

Generally, STUNNING WORK. I had started a react refactor as per #319 but got nowhere near as far as you have here. 

A few questions (mostly ESLint things) but this is really really close to merge. THANK YOU for submitting 😃

public/index.html
rafaelklaessen: PUBLIC_URL is specifically for stuff in the public folder :smile:

src/pages/User/components/PullRequests/ShareButtons.js
rafaelklaessen: What is the benefit of that? It's more difficult to use such a variable, and it isn't available to the frontend.

src/components/PageWrapper.js
rafaelklaessen: We are using Babel, it comes with create-react-app, so don't worry. Without Babel the entire frontend would not work considering that a) JSX is transformed by Babel and not supported by a single browser and b) import/export syntax, or even require, is not supported by browsers without babel.

src/components/UsernameForm/index.js
rafaelklaessen: Set ESLint to support babel-preset-stage-2, it comes with create-react-app by default.

public/index.html
jenkoian: I think we should use the hostname always, some people check out this repo and run it in sub folders and things and can see this potentially being troublesome.

src/components/PageWrapper.js
jenkoian: Sorry should have bene more clear, I meant we aren't using babel explicitly. However, you're right the app wouldn't work without it anyway! In that case, we should probably just use the babel eslint plugin.

src/components/UsernameForm/index.js
jenkoian: Yeah, based off your comments above too I think this is the way to go 👍

public/index.html
rafaelklaessen: As far as I'm aware it's actually a relative path but I'm not sure...
Will change it to REACT_APP_HOSTNAME, as well as replace the use of window.location.origin with REACT_APP_HOSTNAME.

src/pages/User/components/PullRequests/ShareButtons.js
jenkoian: Consistency I guess 🤷‍♂️ it's no biggie. You'll have to excuse my lack of react knowledge here so apologies if this is a bad question, but we're using it here https://github.com/jenkoian/hacktoberfest-checker/pull/333/files#diff-af68e82a3010f5e2e3da304e5fe5caffR45 so could we not pass it through as a prop?

rafaelklaessen:
Some other food for thought, what about 2 spaces rather than 4?

Seems to be the default for the vast majority of React apps, and with 4 spaces everything gets shifted to the right a lot, thereby making code (particularly JSX) more difficult to read. I think...:smile:

src/pages/User/components/PullRequests/ShareButtons.js
rafaelklaessen: No need for passing it through as a prop, it's available everywhere through `process.env.REACT_APP_HOSTNAME`. Will change it :smile:

jenkoian:
@rafaelklaessen yeah thought you might pushback on that 😄 I tend to prefer 4 spaces as I struggle to see the indentation of just 2 😄 perhaps just me or my IDE though and you're right in JS world it does seem that 2 spaces is the norm. If you're happy to update (including .editorconfig and eslint stuff) then happy to go with 2.

rafaelklaessen:
Moved from ESLint to Prettier because it's just so much more simple, and, opposed to ESLint, it just works(TM) with the latest babel presets.
Old eslint commands still work.

package.json
rafaelklaessen: Comes with `react-scripts` :smile:

jenkoian:
Thanks! Almost there. Could you update the 'Running the app' section of the README? If you could do the docker stuff that'd be great but no worries if not.

rafaelklaessen:
Done! :smile: Start command is still the same, but it now uses `parallelshell`.

(Didn't update the docker stuff - I'm not familiar with it.)

jenkoian:
@rafaelklaessen thanks! Gonna give it another look later before merge, but probably the greatest thing I've ever seen 👏

jenkoian:
@rafaelklaessen thanks! Gonna give it another look later before merge, but probably the greatest thing I've ever seen 👏

jenkoian:
@rafaelklaessen thanks! Gonna give it another look later before merge, but probably the greatest thing I've ever seen 👏

jenkoian:
Hey sorry this is taking some long, need to find time to try this out with deployment but time is something in short supply for me right now. Hopefully soon!

jenkoian:
Sorry got no time at mo to run this myself but doing some reading https://facebook.github.io/create-react-app/docs/deployment#other-solutions seems like what we need? Are you able to try this out locally your end and if it works make the required updates?

rafaelklaessen:
So you do want to combine the React and Express server?
I'm sure it will work...I've run create-react-app on Apache 😛

jenkoian:
That seems the easiest thing for me...

jenkoian:
I've pushed a commit which works for me locally. Let me know if you're happy and i'll update the deployment scripts and merge!

rafaelklaessen:
Fixed some tiny issues...There was a space missing in the footer, the build folder wasn't in `.gitignore` and the static serving didn't use an arrow function.

Running both frontend and backend on one server works great.

jenkoian:
Thanks realised it may be worth a note in the .env.example about the port being same on both when running in prod mode too

rafaelklaessen:
I updated .env.example :smile:

jenkoian:
Merged! Thanks so much 👍

PR #361, opened 2019-09-30
Update README.md

Rishabhs343:


jenkoian:
Thanks! I never knew my grammar was so bad at twenty to midnight 😆

PR #369, opened 2019-10-02
Added some hints; suggestion for Message when Event is over

slartibartfass:


src/components/UsernameForm/TimeMessage/getTimeMessage.js
```diff
@@ -6,20 +6,26 @@ const getTimeMessage = () => {
   if (currentMonth !== 9) {
     return "It isn't even October yet!";
   }
+  
+  /*
+  if (daysLeft < 0) {
+    return "Not October anymore. It's all over now! To late to commit your pull requests";
+  }
+  */
```
jenkoian: I'm not gonna merge just the quotes changes, but this bit is interesting. I think having this for a few days after the last day might be nice, but need to make sure it doesn't completely override the message above. So if we can do this but perhaps change it slightly so it only shows up for the first 2 weeks after October or something? We may want to change the wording too, but I'll come back to you on that.

PR #370, opened 2019-10-02
Update .gitignore

FellipeAvanci:


.gitignore
```diff
@@ -7,6 +7,7 @@ node_modules
 
 # Editors
 .idea
+build
```
jenkoian: build is already included, further down.

PR #371, opened 2019-10-02
Add F.A.Q. Page

### LINKED ISSUE #367: Add FAQ section
There are a few things on the app which I think could do with some explaining. For example the new 'pending' text and why some PRs show as counting but the date may be Sept 30th or Nov 1st. 

I think the best thing would be to have a new page called FAQs which answers these curiosities.

Questions I have in mind:

Q: Why do some PRs have 'Pending' next to them?

A: There is a grace period this year which means that a PR must be open (or merged) for at least a week in order to give maintainers the chance to mark issues as invalid. Therefore, the PR will be marked as 'Pending' until that grace period has expired.

Q: Why do some PRs show outside of October?

A: If you've submitted a PR on the last day of September or the first day of November, there is a chance of it counting *if it is October in any timezone*.

--

Are there are any other questions people can think of?

DirtySoc:
Fixes #367

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -9,6 +9,7 @@ import Home from './pages/Home';
 import User from './pages/User';
 import Me from './pages/Me';
 import NotFound from './pages/NotFound';
+import Faq from './pages/Faq'
 
 const App = () => (
   <Fragment>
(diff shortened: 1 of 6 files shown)


DirtySoc:
didn't follow contributing guidelines

PR #377, opened 2019-10-03
fix hacktoberfest label spacing

### LINKED ISSUE #376: Pull request in user page not aligned
Hello!

As can be seen in [this image](https://imgur.com/a/qBm7psP), it looks that some PR entries (maybe all opened PRs?) are unaligned (at least for me).

I tested it in both firefox and chrome, Ubuntu 16.04. My bet is that the div class for it is wrong, as it seems that hacktoberfestp-4 does not exists. I really don't know html and css, so I may be wrong, but I managed to fix it by changing hacktoberfestp-4 to only p-4, as in the others divs.

Is there anyone with the same problem?

vccolombo:
Hello! I opened issue #376 and I think I may found the problem. Patched it in this PR, hope everything is right.

When hacktoberfest label was present in PR, it was adding hacktoberfestp-4 to
div class. This PR fixes this by adding an extra space if 'hacktoberfest' label is present

fix #376

### DIFF
--- a/src/pages/User/components/PullRequests/PullRequest/index.js
+++ b/src/pages/User/components/PullRequests/PullRequest/index.js
@@ -6,7 +6,7 @@ import PullRequestInfo from './PullRequestInfo';
 const PullRequest = ({ pullRequest }) => (
   <div
     className={`bg-white leading-normal ${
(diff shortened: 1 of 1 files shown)


jenkoian:
Thanks @vccolombo

Issue #379, opened 2019-10-03
Glowy title


jenkoian:
I really like the glowy effect on this VS code theme: https://github.com/robb0wen/synthwave-vscode and was wondering if we could do something similar with the `Hacktober` part of the title on the site. 

The CSS from the theme is here: https://github.com/robb0wen/synthwave-vscode/blob/master/synthwave84.css for inspiration

echowebid:
Please review :) #384

jenkoian:
Thanks for all the contributions toward this one, I'm not sure if it actually works on a title at this size, dunno it just looks off to me?

thatzacdavis:
I do think the yellow is the way to go, maybe just more subtle like you said.

PR #381, opened 2019-10-03
Add glow around Hacktober title (Resolves Issue #379)

thatzacdavis:
This is what it looks like with this CSS applied. What do you think?

Screen Shot 2019-10-03 at 1 15 56 PM


jenkoian:
Thanks! Do you think we could make it a bit subtler? And rather than a yellow maybe use a light blue or something?

thatzacdavis:
I used the yellow from the other font to keep is consistent, and spooky 👻 

Could definitely make it more subtle

thatzacdavis:
Screen Shot 2019-10-18 at 11 45 58 PM

How's it looking now?

PR #419, opened 2020-09-30
Adding gzip compression to the server-side

MrLucyfer:
I've added gzip compression to the backend of the app as asked on the issue

### DIFF
--- a/api/index.js
+++ b/api/index.js
@@ -44,8 +44,7 @@ const start = () => {
   app.use(cors(corsOptions));
   
   app.use(compression({
-    filter: shouldCompress,
-    threshold: 0
+    filter: shouldCompress
   }));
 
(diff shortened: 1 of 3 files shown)


api/index.js
```diff
@@ -33,6 +42,11 @@ const start = () => {
   };
 
   app.use(cors(corsOptions));
+  
+  app.use(compression({
+    filter: shouldCompress,
+    threshold: 0
```
jenkoian, line 48: http://expressjs.com/en/resources/middleware/compression.html#threshold do we need this? Should we just leave it to default?

MrLucyfer:
I've read the docs and set the treshold back to default (1kb)

jenkoian (approved):

PR #422, opened 2020-10-01
Improve website footer

willi123yao:
There should be an equal footer margin for both top and bottom, as that will make it look better and balanced on pages where the content is taller than the page height. 

Scrolling to the bottom will make the FAQ button seem too close to the end of the page without the bottom margin.

Also fixed a typo where `class` should be `className` instead.


jenkoian:
Thanks @willi123yao

PR #429, opened 2020-10-02
Exclude draft and spam pull requests

hieplpvip:
According to FAQ, draft and spam pull requests should be ignored as well.

I tried `-label:invalid-label:spam` but it didn't work so I have to use `filter`.


jenkoian:
Brilliant PR, thanks @hieplpvip !

PR #432, opened 2020-10-02
feat: add prettier husky hooks

Ryan-Willpower:
### Why?

I think that formatting the codebase is very important. I've seen the `.prettierrc` file, but some file didn't follow to the prettier standard.

Things that I did:

- Installed packages: husky (for running prettier before commit and push) and prettier-quick (faster write and fix through all staged files).
- Formatted all files that haven't been formatted.
- Removed `eslint` script in package.json because it makes other contributors confuse with actual Eslint.
- Added `--check` argument to the `prettier` script for format checking.

### DIFF
--- a/api/controllers/pr/findPrs/index.js
+++ b/api/controllers/pr/findPrs/index.js
@@ -8,41 +8,47 @@ const loadPrs = require('./loadPrs');
 const findPrs = (github, username) => {
   return loadPrs(github, username)
     .then(pullRequestData =>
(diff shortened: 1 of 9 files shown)


jenkoian:
Thanks, are you able to rebase?

Ryan-Willpower:
Sorry, but I'm outside now. I'll fix it when I'm at home.

jenkoian:
Thanks @Ryan-Willpower

PR #434, opened 2020-10-03
add filter to check if repo has hacktoberfest label

sandeep-chahal:
Not all PR's are eligible for hacktoberfest so I've added a filter to only count the PR's who's repo as hacktoberfest label in it


davwheat:
PRs created before this requirement was introduced (1st and at least some of 2nd) do still count towards your PR total.

jenkoian:
Thanks this has been taken care of in #435 but i've added the label so you get the credit 👍

PR #450, opened 2020-10-06
Move twitter and fb scripts to user page only

Farelion:


src/pages/User/index.js
```diff
@@ -12,6 +12,7 @@ const User = ({
 }) => (
   <Fragment>
     <Helmet>
+      <script src="/components/PullRequests/share.js" type="text/javascript" />
```
jenkoian, line 15: Would be nice to keep this in the footer, would that be possible?

public/index.html
```diff
@@ -74,37 +74,6 @@
...
-
-        t._e = [];
-        t.ready = function(f) {
-          t._e.push(f);
-        };
-        return t;
-      })(document, 'script', 'twitter-wjs');
-    </script>
-    <!-- Facebook -->
     <div id="fb-root"></div>
```
jenkoian, line 77: Would be nice to move this too if poss.

Farelion:
I've made some changes, now fb and twitter scripts that we moved from index.html are inside render of new component called UserShare, that is pulled into user page as last component.
They should show up at the end of body, also moved fb-root as you asked.
Let me know if thats better 😄

Farelion:
Last commit imports UserShare component that i forgot to add oops ;D

src/pages/User/components/PullRequests/UserShare.js
```diff
@@ -0,0 +1,34 @@
+import React, { Component } from 'react';
+
+export default class UserShare extends Component {
```
jenkoian, line 3: Can we use the function method? To be consistent with other components in the app?

`const UserShare = () => (`

Farelion:
Hey, code is working for me, but i can't pass build check and i have no idea why, i'll keep looking.

jenkoian:
@Farelion you can run `npm run prettier-fix` to fix the styling stuff the build is complaining about

Farelion:
After running `prettier-fix` i still couldn't pass test (while uploading with github desktop), i've made some changes to code and it passed (no idea why tho).

jenkoian:
Thanks @Farelion

PR #451, opened 2020-10-06
Add dark/light mode

### LINKED ISSUE #426: Add dark and light mode
The official site lets you switch easily between dark and light mode. This is neat but I think just doing it on users os settings is fine: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme

LMulvey:
Closes #426 

- Used Tailwinds Screens functionality to allow tagging classes for `light-mode`
- Modified `src/style.css` to address a few global styles
- Unrelated to #426 but added a `tailwind-gen-watch` script for running PostCSS in watch mode–works well in conjunction with HMR

I'm not a great designer so by all means if you want a few styling changes, just reach out. Tested in Chrome and Firefox.  Default styles are still "dark" with the media queries exclusively for light mode.

light-mode-3
light-mode-2
light-mode-1


jenkoian:
@LMulvey this is incredible work. Only thing I could find which I think we need to fix is a dark version of the loading animation. https://github.com/jenkoian/hacktoberfest-checker/blob/master/src/pages/User/components/PullRequests/LoadingIcon/loading-icon.gif which is currently white. Are you able to fix?

LMulvey:
@jenkoian Done! 

Sorry, I had been working with the white theme and completely forgot there used to be a loading icon there. I kept the same icon but did some "photoshop magic" to update it for a light theme.

jenkoian:
Thanks @LMulvey looks good, one super minor thing, when I run `npm run tailwind-gen` (not the watch version) I get a slight different `index.css` - just in case it some difference when running `watch` would you mind just re-running `npm run tailwind-gen` to see if you do get different output? It could also be down to differences in node version so make sure you do `nvm use` firs too in case you haven't be.

Anyway super minor thing, amazing work 🤩

LMulvey:
@jenkoian That was 100% it! Forgot to check for a `.nvmrc` before getting started. I've re-generated the CSS with the correct Node version. Hopefully, that will avoid future conflicts on the file.

jenkoian:
Thanks, I still get a slightly different result, but as it's a compiled file, don't worry for now, can fix subsequently - thanks so much for this 🙌

PR #460, opened 2020-10-19
fix hover blue link on light mode

jun-low:
When I hover the register link on the top of the page, it looks like disappear because it's the same color as the background color.
Screenshot 2020-10-18 at 22 01 12

Update with one of the light mode colors in tailwind config:
Screenshot 2020-10-18 at 22 00 58


jenkoian:
Thanks @jun-low !

PR #471, opened 2020-10-27
Fix google fonts not loading space mono

studioromeo:
I noticed we were loading Press start twice but ignoring
space mono. The fallbacks are pretty close so it wasn't
noticable but hey it's worth trying to fix!

I swapped the existing method of loading the fonts to using
links instead which IMO is a bit simpler to use and doesn't
need us to download a javascript file before requesting the
css files to get our font


jenkoian:
Thanks @studioromeo !

PR #475, opened 2020-10-31
chore: Refactor UsernameForm component to function component

MihirGH:
Converted UsernameForm from class to function component

Replaced usage of `withRouter` HOC to `useHistory` hook


jenkoian (approved):
Sorry for the delay but this is great, do you want me to merge or would you rather re-submit in a few days time so it counts towards your 2021 stats?

jenkoian:
Going to merge, thanks for your efforts.

PR #570, opened 2021-09-29
Updated the tailwind css package version

PGNITH:
I've Succesfully Updated the tailwind css package from version 1.9.4 to the current latest version 2.2.16

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "prettier": "^2.1.2",
     "pretty-quick": "^3.1.0",
     "serialize-javascript": "^5.0.1",
-    "tailwindcss": "^1.9.4",
+    "tailwindcss": "^2.2.16",
     "typescript": "^4.0.3"
   },
   "husky": {


jenkoian:
Does it work? Have you tested it? Any new features we unlock? Also, may be worth re-opening in October if you want it to count towards your stats? :)

PGNITH:
Yeah it works fine.
It has a lot of new bug fixes like:
          Improved accessibility of default link focus styles in Firefox 
          Fixed animation variants corrupting keyframes rules
          Ignore escaped commas when splitting selectors to apply prefixes
          Nesting: Maintain PostCSS node sources when handling @apply
          Fix support for animation lists
           Fix arbitrary value support for object-position utilities
          Abort watcher if stdin is closed to avoid zombie processes
          Ignore arbitrary values with unbalanced brackets
And there are much more fixes which the previous version 1.9.4 didn't have.

jenkoian:
@PGNITH thanks, I think there's a few extra things we'd need to take care of:

* We'd need to update the [docker image](https://github.com/jenkoian/hacktoberfest-checker/blob/master/Dockerfile#L1) (and the README) to be node 12 (or poss 14)  https://tailwindcss.com/docs/upgrading-to-v2#upgrade-to-node-js-12-13-or-higher

* Need to make sure our colour palette remains in tact https://tailwindcss.com/docs/upgrading-to-v2#configure-your-color-palette-explicitly

* Check font size remains in tact https://tailwindcss.com/docs/upgrading-to-v2#configure-your-font-size-scale-explicitly

* Ensure postcss and autoprefixer are updated also https://tailwindcss.com/docs/upgrading-to-v2#install-tailwind-css-v2-0-and-post-css-8

* Ensure the `package-lock.json` is updated too

Widi-ps:
why when I write `$ npm run tailwind-gen` it appears like this. Is there something wrong?
image

jenkoian:
@Widi-ps on this branch or master?

Widi-ps:
@jenkoian I think on master, I clone this repo then I started to run the installation but when `$ npm run tailwind-gen` it appears like that. Should I change into branches?

jenkoian:
@Widi-ps it seemed to work for me ok, make sure you're on a supported version of node by using `nvm` (`nvm use`)

Widi-ps:
@jenkoian it still doesn't work, even though I've used nvm lts. is there no error in postcss? or maybe there is another thing that I've missed because at the results something missing with )


image


image

jenkoian:
@Widi-ps I suspect it's a Windows thing or something, this is what it does under the hood:

`node ./node_modules/.bin/postcss ./src/style.css -o ./src/index.css`

So whatever the equivalent of that is that runs ok on your system? I think it could prob be simplified to `postcss ./src/style.css -o ./src/index.css` within `package.json` because npm will add the node_modules to the path implicitly.

jenkoian:
@Widi-ps please try this branch https://github.com/jenkoian/hacktoberfest-checker/pull/594 and if it still doesn't work, try the suggestion in the PR description

PR #583, opened 2021-10-01
Added support for smarter combination of PRs and MRs.

### LINKED ISSUE #576: Smarter combination of PRs and MRs
Currently pull requests from github and merge requests from gitlab and simply concatenated together which means the gitlab ones will just appear after the github ones. What would be better if we merged them smartly so they are combined and ordered by date.

dolamanee6122:
Both PRs and MRs are sorted by date and the most recent ones will come first.

Closes #576

### DIFF
--- a/api/controllers/pr/index.js
+++ b/api/controllers/pr/index.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const moment = require('moment');
 const logCallsRemaining = require('./logCallsRemaining');
 const findGithubPrs = require('./findPrs/github');
 const findGitlabPrs = require('./findPrs/gitlab');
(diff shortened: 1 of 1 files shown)


api/controllers/pr/index.js
```diff
@@ -29,8 +30,15 @@ exports.index = (req, res) => {
         return Promise.reject('notUser');
       }
 
-      // Combine github PRs with the gitlab MRs.
-      prs = prs.concat(mrs);
+      // Combine github PRs with the gitlab MRs in sorted order.
+      // Most recent PRs/MRs will come first.
+      prs = prs.concat(mrs).sort((pr1, pr2) => {
+        const date1 = moment(pr1.created_at, 'MMMM Do YYYY');
+        const date2 = moment(pr2.created_at, 'MMMM Do YYYY');
```
jenkoian, line 37: I was testing this out in jsfiddle https://jsfiddle.net/xzv4a6um/ and I had to remove the `'MMMM Do YYYY'` bit for it to work, otherwise was just getting invalid date. Should we remove it?

dolamanee6122: I looked at the format of `created_at` for PR which looks like "October 1st 2021".
This is because `created_at` is getting formatted to `MMMM Do YYYY` [Here](https://github.com/jenkoian/hacktoberfest-checker/blob/abee8e4d61ce1b26a0bc332ae8b04b8e0ef801bd/api/controllers/pr/findPrs/github/index.js#L45)

```
created_at: moment.utc(event.created_at).format('MMMM Do YYYY')
```
So If I remove  `MMMM Do YYYY`, then the above line also need to be changed to something like this
```
created_at: moment.utc(event.created_at)
```
Then It will work.

jenkoian: @dolamanee6122 Good spot! I forgot we were doing that 😆 - thanks that makes sense 😃

jenkoian:
Thanks!

PR #592, opened 2021-10-03
Fallback for gitlab

babu1998:


babu1998:
Screenshot 2021-10-03 at 9 28 58 PM

For username which is only on gitlab on my localhost

babu1998:
Screenshot 2021-10-03 at 9 29 43 PM

For the username on official hacktoberfest site

babu1998:
have added a lot of if conditions i.e. null checks for many variables and attributes of objects were missing.

api/controllers/pr/index.js
```diff
@@ -23,32 +23,57 @@ exports.index = (req, res) => {
...
     findGitlabPrs(gitlab, username),
-    github.users.getByUsername({ username }).then(logCallsRemaining),
+    searchGithubUser(github, username),
+    searchGitlabUser(gitlab, username),
   ])
-    .then(([prs, mrs, user]) => {
-      if (user.data.type !== 'User') {
+    .then(([prs, mrs, user, gitlab_user]) => {
+      let isGitHubUser = true;
+      if (user == null) isGitHubUser = false;
```
jenkoian, line 31:
```suggestion
      let isGitHubUser = user !== null;
```

api/controllers/pr/index.js
```diff
@@ -23,32 +23,57 @@ exports.index = (req, res) => {
...
+      let isGitLabUser = true;
+      if (JSON.stringify(gitlab_user) == JSON.stringify([]))
```
jenkoian, line 33: How come we comparing json here and not doing the same null check as above?

api/controllers/pr/index.js
```diff
@@ -23,32 +23,57 @@ exports.index = (req, res) => {
...
+        isGitLabUser = false;
+
+      if (
+        (!isGitHubUser || (isGitHubUser && user.data.type !== 'User')) &&
+        !isGitLabUser
+      ) {
```
jenkoian, line 39: I think we can prob get rid of the `user.data.type` check?? Or if we do keep it, do we need a similar one for gitlab?

api/controllers/pr/index.js
```diff
@@ -23,32 +23,57 @@ exports.index = (req, res) => {
...
       // Combine github PRs with the gitlab MRs in sorted order.
       // Most recent PRs/MRs will come first.
-      prs = prs.concat(mrs).sort((pr1, pr2) => {
-        const date1 = moment(pr1.created_at, 'MMMM Do YYYY');
-        const date2 = moment(pr2.created_at, 'MMMM Do YYYY');
-        if (date1.isSame(date2)) return 0;
-        else if (date1.isAfter(date2)) return -1;
-        return 1;
-      });
+      if (prs) {
```
jenkoian, line 45: What about if there are `mrs` but no `prs`??

api/controllers/pr/index.js
```diff
@@ -23,32 +23,57 @@ exports.index = (req, res) => {
...
 
-      const data = {
-        prs,
-        username,
-        userImage: user.data.avatar_url,
-      };
+        res.json(data);
+      } else if (isGitLabUser) {
+        const data = {
+          mrs,
```
jenkoian, line 70: I see why you surrounded the `prs.concat()` thing in an `if` now. Is it necessary? Can we just make `prs` be an empty array if github user doesn't exist and then just concat as usual? Can always just pass `prs` then and simplify this a little so only the `userImage` is conditional, we can use a ternary or something then?

src/pages/User/components/PullRequests/index.js
```diff
@@ -35,24 +35,36 @@ export default function PullRequests({ username }) {
...
     return <ErrorText errorMessage={getErrorMessage()} />;
   }
-
-  const isComplete = data.prs.length >= pullRequestAmount;
+  let isComplete;
+  if (data.prs) {
+    isComplete = data.prs.length >= pullRequestAmount;
+  } else if (data.mrs) {
+    isComplete = data.mrs.length >= pullRequestAmount;
+  }
```
jenkoian, line 43: See note above, but I'm thinking we may not need all the ifs/ternaries in this file.

jenkoian (commented):
Really good work, few comments but seems to work well 👍

api/controllers/pr/index.js
babu1998: gitlab returns [] when no user is found
github returns null when no user is found

api/controllers/pr/index.js
babu1998: we can keep this check assuming github does not change the api. 
This check does not affect gitlab. for gitlab there is separate check !isGitLabUser.

api/controllers/pr/index.js
babu1998: this is a null check

jenkoian (commented):
I think the image size of the gitlab avatars is slightly smaller than the github ones so I _think_ we need to add `class="w-full"` to https://github.com/jenkoian/hacktoberfest-checker/blob/master/src/pages/User/components/PullRequests/UserInfo/UserImage.js#L15 to get rid of that minor black border we get around the user image.

src/pages/User/components/PullRequests/index.js
babu1998: Will think about this. if you have any suggestion let me know.

api/controllers/pr/index.js
jenkoian: Can we normalise upstream? Even if just like `return user_data === null ? [] : user_data` or something?

api/controllers/pr/index.js
babu1998: Nice suggestion. Will try to implement this.

api/controllers/pr/index.js
jenkoian: Yeah, I suppose type _could_ be `Organization` so maybe worth keeping?

api/controllers/pr/index.js
jenkoian: See comment a little further on

src/pages/User/components/PullRequests/index.js
jenkoian: I think if we always assume prs will contains prs _and_ mrs we can just always use `prs` here.

api/controllers/pr/index.js
babu1998: We can. I can make this change for you.

api/controllers/pr/index.js
babu1998: we can remove it since you are planning to add some team/organization features. We dont want to indirectly mess with the logic.

src/pages/User/components/PullRequests/index.js
babu1998: the issue here is in api/controllers/pr/index.js

      if (isGitHubUser) {
        const data = {
          prs,
          username,
          userImage: user.data.avatar_url,
        };

        res.json(data);
      } else if (isGitLabUser) {
        const data = {
          mrs,
          username,
          userImage: user[0].avatar_url,
        };

        res.json(data);
      }

This piece of code changes the attributes of the data object.

jenkoian: What about something like?
```js
  const data = {
    prs,
    username,
    userImage: isGitHubUser ? user.data.avatar_url : user[0].avatar_url,
  };
```

babu1998:
i tested with other usernames in gitlab images are coming out good. The image issue seems to be coming only for that 1 user. The solution of adding class= "w-full" does not work.

babu1998:
Screenshot 2021-10-04 at 6 01 35 PM

babu1998:
Pushed the fixes @jenkoian

api/controllers/pr/index.js
```diff
@@ -28,52 +28,43 @@ exports.index = (req, res) => {
   ])
     .then(([prs, mrs, user, gitlab_user]) => {
       let isGitHubUser = true;
-      if (user == null) isGitHubUser = false;
       let isGitLabUser = true;
+
+      if (JSON.stringify(user) == JSON.stringify([])) isGitHubUser = false;
```
jenkoian, line 33: Do we need to json stringify? Can we just `if (user === [])` ?

api/controllers/pr/index.js
```diff
@@ -28,52 +28,43 @@ exports.index = (req, res) => {
...
-      if (prs) {
-        prs = prs.concat(mrs).sort((pr1, pr2) => {
-          const date1 = moment(pr1.created_at, 'MMMM Do YYYY');
-          const date2 = moment(pr2.created_at, 'MMMM Do YYYY');
-          if (date1.isSame(date2)) return 0;
-          else if (date1.isAfter(date2)) return -1;
-          return 1;
-        });
+      if (prs == null) {
+        prs = [];
```
jenkoian, line 45: 👍

api/controllers/pr/index.js
babu1998: no this does not work in javascript. Because == (and ===) test to see if two objects are the same object and not if they are identical objects.

user and [] are two different objects.

jenkoian: It will be `[]` here: https://github.com/jenkoian/hacktoberfest-checker/pull/592/files#diff-9d1d60f8d642ee863fa2a01bfec91b5440c84bbd481945f1766893c909b7cec3R174 though right? So if we're trying to determine if that github user has returned an error `[] === []` should do it? Or even `user.length` ?

babu1998: [] === [] return false always. You canc check in the console.

babu1998: I can use user.length

jenkoian: > [] === [] return false always. You canc check in the console.
Yeah, I mean if `user` if an empty array, i.e. falls into that error condition.

jenkoian (approved):
Thanks! This is looking really good, will test it locally when I get a sec and merge when happy. Thanks for your efforts :)

Issue #566, opened 2021-09-29
Reskin app


jenkoian:
New year, new colours, should update this app to reflect. Reminder we don't just steal assets or fonts but go for a close match.

danmana:
Hi, I'd like to pick this up.

What should be changed?
* pink -> orange
* logo
* button colors?
* background color?

Also, I'm thinking about how to make this easier to change next year.
Would it be better to define some `hack-primary`, `hack-secondary` colors in tailwind (plus dark mode variants), so we don't have to replace `pink` with `orange` everywhere in HTML/css?
Any feedback is welcome, I'm new to tailwind and I'm not sure what are the best practices.

jenkoian:
Yeah something along the same sort of theme as https://hacktoberfest.digitalocean.com/

Tailwind suggestion sounds good :)

jenkoian:
Closed by #593

PR #593, opened 2021-10-04
Apply 2021 Hacktoberfest theme

nicknamedelta:
First, I would like to thank you and all the other contributors for the successful implementation of this project.

After visiting issue #566, I started working to apply the Hacktoberfest 2021 theme to the project. To do this, I followed the [Hacktoberfest Brandguideline](https://hacktoberfest.digitalocean.com/brandguidelines) which helped me understand how the texts, backgrounds and colors interact.

The result of this work was applied in this PR, however, we may need to refactor the names of the Tailwind CSS variables (I followed @danmana's suggestion in #566 to name variables as e.g. 'hack-context').

I appreciate if you test the modifications and give me some feedback for this PR.


src/components/SiteTitle.js
```diff
@@ -3,13 +3,13 @@ import React from 'react';
 const SiteTitle = () => (
   <div className="text-center">
     <a className="no-underline inline-block py-12 md:py-8" href="/">
-      <h1 className="text-dark-pink text-6xl md:text-7xl">
+      <h1 className="text-hack-logo text-6xl md:text-7xl">
         Hacktober
-        <span className="text-light-pink text-3xl md:text-5xl uppercase block">
+        <span className="text-hack-alt-logo text-3xl md:text-5xl uppercase block">
```
jenkoian, line 8: I think we should remove the uppercase transform here for this year.

src/components/SiteTitle.js
```diff
@@ -3,13 +3,13 @@ import React from 'react';
 const SiteTitle = () => (
   <div className="text-center">
     <a className="no-underline inline-block py-12 md:py-8" href="/">
-      <h1 className="text-dark-pink text-6xl md:text-7xl">
+      <h1 className="text-hack-logo text-6xl md:text-7xl">
         Hacktober
```
jenkoian, line 7: Can we lowercase the `H` now to match this years logo a bit?

src/style.css
```diff
@@ -1,35 +1,34 @@
...
-  @apply bg-dark-blue p-12 text-mid-grey;
+  @apply bg-hack-dark-bg p-12 text-hack-fg;
   font-family: 'Space Mono', monospace;
   padding: 0;
 }
 
 @screen light-mode {
   body {
-    @apply bg-white;
+    @apply bg-hack-fg;
```
jenkoian, line 11: What about using `#f4f0e1` which is the same colour as the background on https://hacktoberfest.digitalocean.com/? It would prob mean a new variable `bg-hack-light` or something. In fact shouldn't this be `bg-hack-alt-bg`?? If so, should we not make that `#f4f0e1` ?

jenkoian: Ah I see `hack-alt-bg` is more the secondary / contrasting colour. So maybe a new var `hack-bg-light` or something?

jenkoian (commented):
Generally looks really good, just a question about the light background colour.

src/style.css
nicknamedelta: For Sure, I created a new variable called '**hack-light-bg**' and applied him to the body of light-mode;

jenkoian:
Thanks, looking better. I think we should remove the alternative font on the `fest` so remove this line https://github.com/jenkoian/hacktoberfest-checker/blob/master/src/style.css#L23 I also think we should tweak the size and spacing too so fest is the same size as the hacktober and sits must closer to it, similar to how it does on https://hacktoberfest.digitalocean.com/ 

That's the only thing, otherwise looks great :)

src/components/SiteTitle.js
```diff
@@ -3,13 +3,13 @@ import React from 'react';
 const SiteTitle = () => (
   <div className="text-center">
     <a className="no-underline inline-block py-12 md:py-8" href="/">
-      <h1 className="text-dark-pink text-6xl md:text-7xl">
-        Hacktober
-        <span className="text-light-pink text-3xl md:text-5xl uppercase block">
+      <h1 className="text-hack-logo text-6xl md:text-7xl">
+        hacktober
+        <span className="block text-hack-alt-logo text-3xl md:text-5xl">
```
jenkoian, line 8: Untested but perhaps `block text-hack-alt-logo text-5xl md:text-6xl -m-8` here to sort the text size and spacing?

nicknamedelta: Changed to `block text-hack-alt-logo text-6xl md:text-7xl -m-8` and worked as well.

src/style.css
```diff
@@ -33,6 +33,8 @@ h1 span {
 
 small {
   font-family: 'Press Start 2P', monospace;
+  font-weight: 600;
+  margin: 20px;
```
jenkoian, line 37: Can we use utilities instead of this? So on the element add `font-semibold m-5` and remove these? That's the last thing, good to go then :)

jenkoian (approved):
Many thanks!

PR #594, opened 2021-10-04
Attempt to fix tailwind-gen for Windows users.

jenkoian:
If this still doesn't work we could try and test `npx tailwindcss build -o src/index.css`


danmana (approved):
Can confirm this works on Windows

Widi-ps:
@jenkoian Yup, this works well on me too

PR #595, opened 2021-10-05
Update latest tailwind, postcss and autoprefixer

Widi-ps:


jenkoian:
Thanks! Can I just check you've run through https://github.com/jenkoian/hacktoberfest-checker/pull/570#issuecomment-931720592 ? The live server is currently on node 10 too, so will need to upgrade that before I merge this.

Widi-ps:
in what version should I upgrade the node? is it the same as me or what?


image

Widi-ps:
I think `lts/*` refers to v14.18.0 because I used the newest in this repo (`nvm use newest`). So I just edit this  `.nvmrc` file with hard code right? like v16.10.0 right?

jenkoian:
@Widi-ps live is now on node 14, so nothing for you to worry about there. For this to progress, are you able to just copy that list from https://github.com/jenkoian/hacktoberfest-checker/pull/570#issuecomment-931720592 and mark each one that you've checked when you've confirmed it's not an issue?

Issue #575, opened 2021-09-30
Gitlab differentiation


jenkoian:
Now that [gitlab support](https://github.com/jenkoian/hacktoberfest-checker/pull/573) has been added find a visual way of differentiating PRs from MRs. Maybe a github and gitlab icon or something?

hwilson2563:
I would like to come up with a solution for this issue.

PR #596, opened 2021-10-05
add: icons based off url

hwilson2563:
I used PNGs of the different platforms' logos. I wasn't sure if I should add a background color div to make the black outlines of the PNGs stand out since the pr for the new theme isn't merged I thought maybe this could be addressed following that merge. 
This is to resolve issue: Gitlab differentiation #575

Here are screenshots: 
image
image


jenkoian:
This looks pretty cool! Thanks! It does look slightly better with the old design, but still think it works with the new. I'll approve anyway but let me know if you want to tweak now the new design is live, otherwise I'm happy to just merge :)

jenkoian (approved):

jenkoian:
@hwilson2563 I believe it will count as the PR has been approved, but will add it anyway :)

jenkoian:
Only the github icon on dark mode seem to change for me? I reckon just use the light version in all instances?

PR #598, opened 2021-10-06
social media added

Ttecs:
added social media buttons


jenkoian:
The links don't seem to go anywhere? Besides we already have social icons after you submit a username, I'm not sure it's worth having them before submission tbh, and even if so I think we should re-use the existing icons rather than creating new.

Thanks for the PR but it's a no on this I'm afraid.

PR #602, opened 2021-10-10
Remove unneeded Fragments

AlexMayol:
There's a [shorter syntax](https://reactjs.org/docs/fragments.html#short-syntax) to write Fragments that do not require keys or attributes, like the ones used in this codebase. 

Also, some Tailwind classes have beer reordered automatically in those files with the [Headwind extension for VSCode](https://marketplace.visualstudio.com/items?itemName=heybourn.headwind), a very useful one.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,4 +1,4 @@
-import React, { Fragment } from 'react';
+import React from 'react';
 import { Helmet } from 'react-helmet';
 import { BrowserRouter as Router, Switch, Route } from 'react-router-dom';
 import GithubCorner from './components/GithubCorner';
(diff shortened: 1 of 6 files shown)


jenkoian:
Happy with the fragment change, do you know what the justification is for the default order in Headwind? I had a look at the extension page you linked to but couldn't find an answer? I'm generally 👎 on enforcing an order for css rules like this unless there's a legit reasoning. I also wouldn't want to force something like Headwind on people, as not everyone uses VSCode.

AlexMayol:
@jenkoian I believe Headwind orders classes as padding/margins > height/widths > any breakpoint modifier. 
I didn't include anything to force Headwind, I just wanted to note why the classes were changed in addition to the Fragments.
Do you want me to remove the change to the classes?

jenkoian:
@AlexMayol it's ok, I don't think it will conflict with any open PRs so happy to leave it in.

PR #603, opened 2021-10-10
Refined imports in components and pages

DheekshaGopika:
Created an index.js on both `pages/` and `components/` folders for easy importing components on other files

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -5,7 +5,7 @@ import {
   GithubCorner,
   RegisterReminder,
   PageWrapper,
-  Footer
+  Footer,
 } from './components';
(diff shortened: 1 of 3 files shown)


jenkoian:
@DheekshaGopika I don't work too much with JS/React - is this kind of thing considered good practice?

AlexMayol:
I also love the concept of [absolute imports](https://medium.com/hackernoon/absolute-imports-with-create-react-app-4c6cfb66c35d) and I use them in my NextJS projects, would this be a good addition? @jenkoian

jenkoian:
@AlexMayol yeah that looks cool

PR #613, opened 2021-10-22
FAQ page footer's FAQ button changed to Home button

developer-nobita:
Fixes #601 
Changed footer component to treat FAQ page and other pages differently.

Screenshots:

Screenshot (33)

Screenshot (34)

PR #614, opened 2021-10-23
Make FAQ link conditional

aarai:
Remove FAQ link from FAQ page

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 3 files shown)


jenkoian:
Is this just a slightly different take on https://github.com/jenkoian/hacktoberfest-checker/pull/613 ? Would rather just stick to the one approach so will likely close this unless I'm misunderstanding?

aarai:
Hi no you’re correct. 613 is the one that should be accepted. My PR should be closed. 

Thanks

Issue #705, opened 2022-10-01
Invisible links


Ernest2026:
Some link on the page are invisible. It has the same color as the background in light mode.

E.g: "Disclaimer: This site is fan made and not affiliated with __[Hacktoberfest](https://hacktoberfest.digitalocean.com/)__."

darshikaguptaa:
Can you assign this issue to me ?

Ernest2026:
Sure, no problem

DevMukhtarr:
I found a link that isn't visible when you hover over it asides from the invisible link @darshikaguptaa is handling
<img width="877" alt="Screenshot (4)" src="https://user-images.githubusercontent.com/50122611/194150130-6259efa9-2608-4e51-8fa7-a7fa9617e453.png">

PR #706, opened 2022-10-02
Invisible links issue solved

darshikaguptaa:
Issue #705 Solved

### DIFF
--- a/src/components/Footer.js
+++ b/src/components/Footer.js
@@ -8,6 +8,7 @@ const Footer = () => (
     <p className="text-hack-fg light-mode:text-hack-dark-title">
       Disclaimer: This site is fan made and not affiliated with{' '}
       <a
+      style={{color:"black"}}
         href="https://hacktoberfest.digitalocean.com/"
         target="_blank"
         rel="noopener noreferrer"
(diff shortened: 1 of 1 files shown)


src/components/Footer.js
```diff
@@ -8,6 +8,7 @@ const Footer = () => (
     <p className="text-hack-fg light-mode:text-hack-dark-title">
       Disclaimer: This site is fan made and not affiliated with{' '}
       <a
+      style={{color:"black"}}
```
jenkoian, line 11: We use tailwind for styling, can you change this to a className attribute instead? Also, ideally want to re-use the colours declared in the tailwind config file. The colours are a bit of a mess and need sorting out, but in the mean time, you can probably use one of the colours there. See the surrounding p tag here for an example of some text colours and also the light-mode modifier.

darshikaguptaa: Sure I'll try to fix it

darshikaguptaa:
Hey @jenkoian  I've changed the code according to tailwind config file

vimode:
The same fix can be applied to links on hover.


As of now the hover effect too blends in with the background.

darshikaguptaa:
@jenkoian are there any changes to be made?

darshikaguptaa:
@jenkoian I've added the hover modifier as well

jenkoian:
Thanks @darshikaguptaa

PR #711, opened 2022-10-05
bugfix: remove whitespaces from username input field

### LINKED ISSUE #709: Input validation required
Hello,

If you search a username with an empty space before the username, it will throw an error. 
I think an input validation should be added on submit which will make sure that any whitespace before the input value is removed. 

AFAIK, github usernames cannot have spaces before anyway, so that shouldn't be a problem either. Isn't it?

vimode:
Possible solution to fix #709 

Do you think this solution is a little user hostile? 
Would it be better to apply the trim() method only onSubmit to all the usages of state `username` thereafter? 

Thank you.

### DIFF
--- a/src/components/UsernameForm/index.js
+++ b/src/components/UsernameForm/index.js
@@ -15,7 +15,7 @@ export default function UsernameForm(props) {
   const navigate = useNavigate();
 
   const handleUsernameChange = useCallback(
-    (e) => setUsername(e.target.value.trim()),
+    (e) => setUsername(e.target.value),
     []
   );
 
@@ -28,9 +28,9 @@ export default function UsernameForm(props) {
       }
 
       if (props.onCheckUser) {
-        props.onCheckUser(username);
+        props.onCheckUser(username.trim());
       } else {
-        const userUrl = getUserUrl(username);
+        const userUrl = getUserUrl(username.trim());
         navigate(userUrl);
       }
     },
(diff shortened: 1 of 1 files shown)


jenkoian:
Oh you mean because this means you can't enter spaces at all...hmm yeah I think onSubmit would be nicer if easy enough? Don't think it's too bad though if it is hard.

vimode:
I agree it isn't very bad as spaces aren't a valid character in usernames on either of the websites. 
It just feels a little aggressive 😄 

Let me try a few things. 🤔 

- Instead of using trim(), let me try using a regex pattern to replace the the whitespace with "", something like  `e.target.value.replace(/\s+/ig, "")`, if it has a little different and has a less aggressive behavior then that would be better. 

- If regex doesn't work then, I would use trim() onSubmit to instances of `username` being passed, as it looks like `username` is passed as an argument to `onCheckUser` and `getUserUrl`  and nowhere else from this component. 

I will get back to you as soon as I try these possible solutions, in the meantime if you have other ideas or suggestions or advice or tips I am open to all of those 😃

vimode:
So the regex replace works the same as trim() if used with onChange. 

So I used the trim while passing the state `username` as an argument in the onSubmit function. Everything should work fine. 🤞

jenkoian:
Thanks @vimode

PR #723, opened 2022-10-10
Made search bar visible without scrolling 

### LINKED ISSUE #719: The search bar is too down
When we enter the website, we cannot directly see the search bar. I can make UX better by making the search bar appear on the screen directly when we enter the website without scrolling.
#### As you see we need to scroll to see the search bar
Please assign this to me and I will make the styling changes 

image

Saran-pariyar:
# I did some styling changes so that we could we the search bar directly when we open the website without scrolling.

###  Before: 
Screenshot from 2022-10-10 08-11-03

### Now : 
Screenshot from 2022-10-10 08-07-59

Sorry but I'm new and didn't understand how to close the issue, but this commit solves the issue of  no #719

### DIFF
--- a/src/components/Navbar.js
+++ b/src/components/Navbar.js
@@ -20,7 +20,7 @@ const Navbar = () => {
 
   return (
     <>
-      <nav className="text-center pb-4">
+      <nav className="text-center pb-4 nav">
         {me && (
           <NavLink
             to="/me"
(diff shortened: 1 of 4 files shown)


vimode:
Hey @Saran-pariyar  
Please read this guideline
https://github.com/jenkoian/hacktoberfest-checker#changing-styling

>Styling is done through tailwind as much as possible, no CSS should be added and class names favoured. However, there will be occasions where custom CSS is required, in this case you need to edit style.css and run npm run tailwind-gen to generate the new CSS file.
Note: You should never directly edit index.css

Try using tailwind as much as possible.

PR #729, opened 2022-10-15
Update style.css

elite16op:
did changes on styles.css you may check it now, to solve that link fade issue


elite16op:
> did changes on styles.css you may check it now, to solve that link fade issue  

Owner please update me the problem that why haven't you responded to the pr.

PR #735, opened 2022-10-20
Updating gitignore

pragapraga:


jenkoian:
Thanks but I'm of the belief people should set these up in their global gitignore rather than on a project level.

PR #739, opened 2022-10-22
Change Github Corner

Aniketmishra0:
## Change Github Corner

Screenshot 2022-10-22 at 11 18 02 AM


jenkoian:
Thanks but I think the existing is more inline with this years theme.

PR #740, opened 2022-10-22
show live remaining time

### LINKED ISSUE #736: Show Live remaining time left instead of just Days
The site currently just shows the remaining days of hacktoberfest. It will be nice to update live ( days + time ) every second.

I would like to work on it, please assign it to me.

akankshasainics:
Close #736

### DIFF
--- a/src/components/UsernameForm/TimeMessage/getTimeMessage.js
+++ b/src/components/UsernameForm/TimeMessage/getTimeMessage.js
@@ -45,7 +45,7 @@ class RemainTime extends React.Component {
       <div className="time">
         {timeMessage}
         <br />
-        {remainTime}
+        {this.state.days != null ? remainTime : ''}
       </div>
     );
   }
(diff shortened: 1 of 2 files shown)


jenkoian:
Thanks, I like it, just a few things:

* When the page first loads or you navigate to another page it displays `null:null:null:null` in place of the time. Can we just hide it until the time is loaded or something?
* As it's client side I _think_ it's ok to just use midnight on 01/11 to end the timer, but just a point to consider timezones (but I think we're ok as mentioned)

akankshasainics:
@jenkoian I have added a new [commit](https://github.com/jenkoian/hacktoberfest-checker/pull/740/commits/72f885978fb8e9b65d568cb2500263ad9b02728c) to fix 1st problem. Can you please explain a little more about the 2nd? What changes do I have to make? By the way, the Code will end the timer on 01/11 at midnight.

jenkoian:
Thanks @akankshasainics

